PR #1, opened 2015-09-22
do not break ancestry tree

seletskiy:
Running `mark` for updating page, deeply nested in some tree, ended up page
being shifted to the Ctulhu's void, e.g. not belonging to any parent.

### DIFF
--- a/main.go
+++ b/main.go
@@ -56,10 +56,15 @@ Options:
 )
 
 type PageInfo struct {
-	Title   string `json:"title"`
+	Title string `json:"title"`
+
 	Version struct {
 		Number int64 `json:"number"`
 	} `json:"version"`
+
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -188,6 +193,9 @@ func updatePage(
 			"number":    nextPageVersion,
 			"minorEdit": false,
 		},
+		"ancestors": []map[string]interface{}{
+			{"id": pageInfo.Ancestors[len(pageInfo.Ancestors)-1].Id},
```
kovetskiy: please move this stuff to variable like `nextPageVersion` and check for len(pageInfo.Ancestors) > 0

kovetskiy:
thank you for contribution

PR #4, opened 2019-02-06
change page title

brnv:


kovetskiy:
I'm not sure why we would need it. If we are going for DaaC (Documentation
as a Code) we don't want to store some external ID in source documentation.

brnv:
up

main.go
```diff
@@ -285,7 +285,40 @@ func main() {
 
 	var target *PageInfo
 
-	if meta != nil {
+	var rename *PageInfo
+
+	if len(meta.Title) > 1 {
+		page, _ := api.findPage(meta.Space, meta.Title[len(meta.Title)-1])
+		if page == nil {
+			for _, title := range meta.Title {
```
kovetskiy: we need to go in reversed order

brnv: check pls

PR #6, opened 2019-07-23
api: use different restrict api for cloud

seletskiy:
Also, fix help message.


seletskiy:
pik balmera :cocktail:

seletskiy:
@kovetskiy zamerszhi uzhe

PR #8, opened 2019-08-06
implement macros & includes

seletskiy:


README.md
```diff
@@ -30,10 +30,127 @@ File in extended format should follow specification
...
+be replaced with specified template:
+
+```markdown
+<!-- Macro: <regexp>
+     Template: <path>
+     <yaml-data> -->
+```
+
+Capture groups can be defined in the macro's `<regexp>` which can be later
+referenced in the `<yaml-data>` using `${<number>}` syntax.
```
kovetskiy: Didn't get, what the `<number>` here? It's mentioned here first time

main.go
```diff
@@ -117,50 +137,85 @@ func main() {
...
-	if dryRun {
-		fmt.Println(string(mark.CompileMarkdown(markdownData)))
-		os.Exit(0)
+	meta, err := mark.ExtractMeta(markdown)
+	if err != nil {
+		log.Fatal(err)
 	}
 
-	creds, err := GetCredentials(args, config)
+	stdlib, err := stdlib.New(api)
```
kovetskiy: can we do something less abstract? maybe longer name but more obvious about purposes of the package

pkg/mark/includes/templates.go
```diff
@@ -0,0 +1,155 @@
...
+	"text/template"
+
+	"gopkg.in/yaml.v2"
+
+	"github.com/kovetskiy/mark/pkg/log"
+	"github.com/reconquest/karma-go"
+)
+
+var (
+	reIncludeDirective = regexp.MustCompile(`(?s)<!-- Include: (\S+)(.*?)-->`)
```
kovetskiy: The first space can be omitted too in classic HTML
`<!--\s*Include:`

pkg/mark/macro/macro.go
```diff
@@ -0,0 +1,162 @@
...
+	"text/template"
+
+	"github.com/kovetskiy/mark/pkg/log"
+	"github.com/kovetskiy/mark/pkg/mark/includes"
+	"github.com/reconquest/karma-go"
+	"gopkg.in/yaml.v2"
+)
+
+var reMacroDirective = regexp.MustCompile(
+	`(?s)<!-- Macro: ([^\n]+)\n\s*Template: (\S+)\n(.*?)-->`,
```
kovetskiy: the first space can be omitted here too

pkg/mark/macro/macro.go
```diff
@@ -0,0 +1,162 @@
...
+		contents,
+		func(spec []byte) []byte {
+			if err != nil {
+				return spec
+			}
+
+			groups := reMacroDirective.FindSubmatch(spec)
+
+			var (
+				expr, path, config = groups[1], string(groups[2]), groups[3]
```
kovetskiy: please use regexputil-go to avoid this ambiguous indexing

pkg/mark/includes/templates.go
```diff
@@ -0,0 +1,155 @@
...
+)
+
+func LoadTemplate(
+	path string,
+	templates *template.Template,
+) (string, *template.Template, error) {
```
kovetskiy: that returned `string` which is name is unused in the code

pkg/mark/stdlib/stdlib.go
```diff
@@ -0,0 +1,153 @@
...
+			/**/ `<ac:layout>`,
+			/**/ `<ac:layout-section ac:type="two_right_sidebar">`,
+			/**/ `<ac:layout-cell>{{ .Body }}</ac:layout-cell>`,
+			/**/ `<ac:layout-cell></ac:layout-cell>`,
+			/**/ `</ac:layout-section>`,
+			/**/ `</ac:layout>`,
+			`{{ else }}`,
+			/**/ `{{ .Body }}`,
+			`{{ end }}`,
+		),
```
kovetskiy: just a quick note: we could also store these templates in files and distribute them along with binary

main.go
seletskiy: I can't come up with a better name. Do you have some variants in mind?

pkg/mark/stdlib/stdlib.go
seletskiy: Yep, but on the other hand it will make tool not ready to use after `go get`/`go install`.

PR #12, opened 2019-11-22
Fixed #11: Allow Confluence Space to be mixed case

### LINKED ISSUE #11: Allow mixed case Space names
The instance of confluence I use allows mixed case Space names.  https://github.com/kovetskiy/mark/blob/master/pkg/mark/meta.go#L81 changed to `meta.Space = strings.TrimSpace(value)` works for me.  

I poked around (a little bit) in the Confluence documentation and couldn't find anything that suggested Space names can only be upper case.

dauc:
Swap `ToUpper` for `TrimSpace` in meta parser to leave the case as it
was passed.


kovetskiy:
Thanks!

PR #18, opened 2020-03-06
Add Go Module Support

klauern:
This adds support for Go modules.  Not a lot to change to make it work, which is
nice.


kovetskiy:
Thanks

PR #22, opened 2020-06-25
Add ExactAttachment header

csoutherland:
Added an `ExactAttachment` header that doesn't require the use of an `attachment://` URL scheme. This should simplify using `mark` as a tool for something like keeping GitHub and Confluence in sync at the cost of adding a tool that may give undesired behavior in some circumstances if used without caution.

Unfortunately, I had to change the `pkg/mark/Meta` data structure to get an approach that was relatively easy on the eyes. However, I figured it might be a good place to make the change in case a syntactic change is preferable over `ExactAttachment` and/or regular expression replacement becomes a priority.

I tried both attachment mechanisms with my local build and everything _seems_ to be working with the refactor and the new header. In fact, I now have a GitHub and Confluence page that look more or less identical (including locally generated images) from the same markdown code, which was at least my own goal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ Confluence), so `ExactAttachment` has been introduced:
 An attached link is [here](<path>)
 ```
 
-*NOTE* Be careful with `ExactAttachment`! If your path string is a subset of
+**NOTE**: Be careful with `ExactAttachment`! If your path string is a subset of
 another longer string or referenced in text, you may get undesired behavior.
 
 Mark also supports macro definitions, which are defined as regexps which will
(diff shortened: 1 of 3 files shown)


kovetskiy:
Thanks for the PR! 

I've been thinking about this feature for a few days, what if we actually drop current `Attachment` implementation and use the ExactAttachment implementation instead of it?

But in order to save backward compatibility we will have to look for attachment:// links first.

Like as the following:
1. Parsed `<-- Attachment: x.png -->`
2. Find attachment://x.png in markdown, if found, then replace and go to step 4, if not go to step 3
3. Find x.png in markdown, replace if found 
4. Done

What do you think?

csoutherland:
Oh, sorry, got pulled into some other things and never noticed the update on this PR. I really like the approach you ended up with!

PR #24, opened 2020-07-29
Update Readme for new Docker usage

tillepille:
Since you patched your Entrypoint to a shell instead of directly calling `mark` the documentation should reflect that.


kovetskiy:
Thanks!

PR #33, opened 2020-11-12
Support for relative links

### LINKED ISSUE #25: Relative links support
Hi team,
And thank you for the awesome tool!
Do you have any plans to support relative link in articles? Or maybe it is already there, but I didn't get it?

Right now such type of links:
`[Configuration](docs/configuration.md)`
turns into 
`(https://<confluence_link>/display/<space_name>/docs/configuration.md)`

Hi-Fi:
Fixes #25

### DIFF
--- a/main.go
+++ b/main.go
@@ -212,6 +212,12 @@ func main() {
 		}
 	}
 
+	links, err := mark.ResolveRelativeLinks(api, markdown, ".")
+	if err != nil {
+		log.Fatalf(err, "unable to resolve relative links")
+	}
+	markdown = mark.ReplaceRelativeLinks(markdown, links)
+
 	if dryRun {
 		compileOnly = true
 
--- a/pkg/mark/link.go
+++ b/pkg/mark/link.go
@@ -25,36 +25,47 @@ func ResolveRelativeLinks(
 	markdown []byte,
 	base string,
 ) (links []Link, collectedErrors error) {
-	re := regexp.MustCompile("\\[.*?\\]\\((.*?)\\)")
-	submatchall := re.FindAllStringSubmatch(string(markdown), -1)
-	for _, element := range submatchall {
-		possibleMDFile := element[1]
-		filepath := filepath.Join(base, possibleMDFile)
-		if _, err := os.Stat(filepath); err == nil {
-			linkMarkdown, err := ioutil.ReadFile(filepath)
-			if err != nil {
-				collectedErrors = fmt.Errorf("%w\n unable to read markdown file "+filepath, collectedErrors)
-				continue
-			}
-			meta, _, err := ExtractMeta(linkMarkdown)
-			if err != nil {
-				collectedErrors = fmt.Errorf("%w\n unable to get metadata from markdown file "+filepath, collectedErrors)
-				continue
-			}
-			link := fmt.Sprintf("%s/display/%s/%s", api.BaseURL, meta.Space, url.QueryEscape(meta.Title))
-			confluencePage, err := api.FindPage(meta.Space, meta.Title)
-			if err != nil {
-				collectedErrors = fmt.Errorf("%w\n "+err.Error(), collectedErrors)
-			} else if confluencePage != nil {
-				// Needs baseURL, as REST api response URL doesn't contain subpath ir confluence is server from that
-				link = api.BaseURL + confluencePage.Links.Full
-			}
+	currentMarkdownMetadata, onlyMarkdown, err := ExtractMeta(markdown)
+	if err != nil {
+		return links, fmt.Errorf("unable to get metadata from handled markdown file. Error %w", err)
+	}
 
-			links = append(links, Link{
-				MDLink: possibleMDFile,
-				Link:   link,
-			})
+	currentPageLinkString, collectedErrors := getConfluenceLink(api, currentMarkdownMetadata.Space, currentMarkdownMetadata.Title, collectedErrors)
+
+	submatchall := collectLinksFromMarkdown(string(onlyMarkdown))
+
+	for _, element := range submatchall {
+		link := Link{
+			MDLink: element[1],
+			Link:   currentPageLinkString,
 		}
+		// If link points to markdown like target, we build link for that in Confluence
+		if len(element[2]) > 0 {
+			possibleMDFile := element[2]
+			filepath := filepath.Join(base, possibleMDFile)
+			if _, err := os.Stat(filepath); err == nil {
+				linkMarkdown, err := ioutil.ReadFile(filepath)
+				if err != nil {
+					collectedErrors = fmt.Errorf("%w\n unable to read markdown file "+filepath, collectedErrors)
+					continue
+				}
+				// This helps to determine if found link points to file that's not markdown
+				// or have mark required metadata
+				meta, _, err := ExtractMeta(linkMarkdown)
+				if err != nil {
+					collectedErrors = fmt.Errorf("%w\n unable to get metadata from markdown file "+filepath, collectedErrors)
+					continue
+				}
+
+				link.Link, collectedErrors = getConfluenceLink(api, meta.Space, meta.Title, collectedErrors)
+			}
+		}
+
+		if len(element[3]) > 0 {
+			link.Link = currentPageLinkString + "#" + element[2]
+		}
+
+		links = append(links, link)
 	}
 	return links, collectedErrors
 }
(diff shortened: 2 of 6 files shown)


main.go
```diff
@@ -212,6 +212,9 @@ func main() {
 		}
 	}
 
+	links, err := mark.ResolveRelativeLinks(api, markdown, ".")
```
kovetskiy, line 215: You forgot to handle an error here.

pkg/mark/link.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"github.com/kovetskiy/mark/pkg/confluence"
+)
+
+type Link struct {
+	MDLink string
+	Link   string
+}
+
+// ResolveRelativeLinks finds links in the markdown, and replaces one pointing to other Markdowns either by own link (if not created to Confluence yet) or witn actual Confluence link
```
kovetskiy, line 20: Could you wrap the comment to 80 characters per line? This long is too long, even GitHub can't fit it in the screen size.

pkg/mark/link.go
```diff
@@ -0,0 +1,69 @@
...
+	)
+	re := regexp.MustCompile("\\[.*?\\]\\((.*?)\\)")
+	submatchall := re.FindAllStringSubmatch(string(markdown), -1)
+	for _, element := range submatchall {
+		possibleMDFile := element[1]
+		filepath := filepath.Join(base, possibleMDFile)
+		if _, err := os.Stat(filepath); err == nil {
+			linkMarkdown, err := ioutil.ReadFile(filepath)
+			if err != nil {
+				continue
```
kovetskiy, line 37: Errors here and after should be reported to the user. It's not fine that mark can't read a file.

pkg/mark/link.go
```diff
@@ -0,0 +1,69 @@
...
+			}
+			meta, _, err := ExtractMeta(linkMarkdown)
+			if err != nil {
+				continue
+			}
+			link := fmt.Sprintf("%s/display/%s/%s", api.BaseURL, meta.Space, url.QueryEscape(meta.Title))
+			confluencePage, err := api.FindPage(meta.Space, meta.Title)
+			if err == nil && confluencePage != nil {
+				// Needs baseURL, as REST api response URL doesn't contain subpath ir confluence is server from that
+				link = fmt.Sprintf("%s%s", api.BaseURL, confluencePage.Links.Full)
```
kovetskiy, line 47: No need for fmt.Sprintf here, just concatenate these two strings with `+`.

pkg/mark/link.go
```diff
@@ -0,0 +1,69 @@
...
+			linkMarkdown, err := ioutil.ReadFile(filepath)
+			if err != nil {
+				continue
+			}
+			meta, _, err := ExtractMeta(linkMarkdown)
+			if err != nil {
+				continue
+			}
+			link := fmt.Sprintf("%s/display/%s/%s", api.BaseURL, meta.Space, url.QueryEscape(meta.Title))
+			confluencePage, err := api.FindPage(meta.Space, meta.Title)
```
kovetskiy, line 44: Don't ignore the error, report it to the user.

pkg/mark/link.go
```diff
@@ -0,0 +1,69 @@
...
+	Link   string
+}
+
+// ResolveRelativeLinks finds links in the markdown, and replaces one pointing to other Markdowns either by own link (if not created to Confluence yet) or witn actual Confluence link
+func ResolveRelativeLinks(
+	api *confluence.API,
+	markdown []byte,
+	base string,
+) ([]Link, error) {
+	var (
```
kovetskiy, line 26: No need for group declaration if you are not grouping them, you can assign links to an empty slice:
```
links := []Link{}
```

pkg/mark/link.go
```diff
@@ -0,0 +1,69 @@
...
+		links []Link
+	)
+	re := regexp.MustCompile("\\[.*?\\]\\((.*?)\\)")
```
kovetskiy, line 29: Isn't this regexp too greedy? I see you are using `.*`

kovetskiy (changes requested):
Thanks for the contribution, there are several things we need to address before the merge.

pkg/mark/link.go
Hi-Fi: It could be made a bit less greedier, but I don't know it it's reasonable to modify as it still stops at first closing parenthesis (both normal and square). 

FIrst part in square brackets can be "anything" (as it's just text), so only latter part is something that might have better regexps.

main.go
Hi-Fi: Added

pkg/mark/link.go
Hi-Fi: Logging errors, but not breaking.

pkg/mark/link.go
Hi-Fi: Wrapped those long comments.

pkg/mark/link.go
```diff
@@ -0,0 +1,73 @@
...
+	links := []Link{}
+	re := regexp.MustCompile("\\[.*?\\]\\((.*?)\\)")
+	submatchall := re.FindAllStringSubmatch(string(markdown), -1)
+	for _, element := range submatchall {
+		possibleMDFile := element[1]
+		filepath := filepath.Join(base, possibleMDFile)
+		if _, err := os.Stat(filepath); err == nil {
+			linkMarkdown, err := ioutil.ReadFile(filepath)
+			if err != nil {
+				log.Errorf(err, "unable to read markdown file %s", filepath)
```
kovetskiy, line 38: This is much better now, but don't forget that the function has `([]Link, error)` returning signature, so we don't need to log errors here, we should return the first occurred error. 

```
if err != nil
    return nil, karma.Format(err, "unable to read markdown file: %s", filepath)
}
```

pkg/mark/link.go
kovetskiy: Oops, you resolved the conversation but forgot to handle err returned by api.FindPage

kovetskiy (changes requested):

pkg/mark/link.go
Hi-Fi: Hmm... maybe then some list of failing ones should be made, as I don't think it's reasonable to end if first one finds error. So no return in the loop, but after it

pkg/mark/link.go
Hi-Fi: Hmm... there could be some print that api call failed, but at the end the default value is put in that case to the conversion so it's not fatal (same as if page is not created yet to Confluence).

pkg/mark/link.go
```diff
@@ -0,0 +1,73 @@
...
+
+// ResolveRelativeLinks finds links in the markdown, and replaces one pointing
+// to other Markdowns either by own link (if not created to Confluence yet) or
+// witn actual Confluence link
+func ResolveRelativeLinks(
+	api *confluence.API,
+	markdown []byte,
+	base string,
+) (links []Link, collectedErrors error) {
+	re := regexp.MustCompile("\\[.*?\\]\\((.*?)\\)")
```
lukiffer, line 28: Where links contain anchors (e.g. `../path/to/example.md#second-heading`), this pattern treats that as part of the filename in `ioutil.ReadFile()` below. Moreover, we should preserve these anchors as they're (mostly) compatible between readers.

Consider using `\[[^\]]+\]\(([^\)#]+)?#?(.*)?\)` instead (it's a bit more efficient and will capture both parts of the URL.

```
[example](../path/to/example.md#second-heading)
$1 = "../path/to/example/md"
$2 = "second-heading"

[example](../path/to/example.md)
$1 = "../path/to/example.md
$2 = ""

[example](#heading-in-document)
$1 = ""
$2 = "heading-in-document"
```

lukiffer: Then below: 
- when both `$1` and `$2` are present, append `#$2` to the resolved link.
- when `$1` is empty, prevent the protocol prefix and instead use `#$2` verbatim.

lukiffer: After some further digging, it appears Confluence doesn't like `#anchor` links alone. Given that it's possible this page hasn't been created yet, we can't deterministically query for its Confluence URL.

I suppose the same is true depending on the order pages are created. (If you create two new pages, A and B, and A links to B, and B links to A, you wouldn't have links working correctly on all pages until after two deployment passes).

I suppose because the tool is designed to run on a single file, this is probably by design, and it puts the responsibility of mapping this on the user.

We could attempt to resolve the _current_ page and link to itself (with anchor) as a fully qualified URL.

Hi-Fi: Thank you, will check the link finding regexp and logic.

Hi-Fi: My implementation handles that "not created page" by self building the URL that page is going to get (works at least with ASCII characters, not sure about UTF8). If page is created already it's URL is fetched from confluence.

With anchor alone I think one way is to get current page's (coming) URL, and append anchor there.

Hi-Fi: Added third group to regexp, as whole content from inside parenthesis is needed when making the actual replace.

kovetskiy:
Thank you for the contribution and your patience @Hi-Fi 
Thank you for the help with the code review @lukiffer

PR #36, opened 2020-11-17
Add Option for Dropping Leading H1 Heading

lukiffer:
This functionality is designed to eliminate duplicate page titles in Confluence (once from the page title itself, and a second from the rendered H1), but preserve them for display in GitHub or other non-contextual environments.

- Adds a `--drop-h1` flag that removes leading ATX H1 headings, excluding them from the Confluence wiki markup output.
- Adds `.idea/` (GoLand) IDE preferences to `.gitignore`

### DIFF
--- a/main.go
+++ b/main.go
@@ -108,7 +108,6 @@ By default, mark provides several built-in templates and macros:
 Usage:
   mark [options] [-u <username>] [-p <token>] [-k] [-l <url>] -f <file>
   mark [options] [-u <username>] [-p <password>] [-k] [-b <url>] -f <file>
-  mark [options] [-u <username>] [-p <password>] [-k] [-n] -c <file>
   mark -v | --version
   mark -h | --help
 
--- a/main.go
+++ b/main.go
@@ -108,7 +108,6 @@ By default, mark provides several built-in templates and macros:
 Usage:
   mark [options] [-u <username>] [-p <token>] [-k] [-l <url>] -f <file>
   mark [options] [-u <username>] [-p <password>] [-k] [-b <url>] -f <file>
-  mark [options] [-u <username>] [-p <password>] [--drop-h1] -f <file>
   mark -v | --version
   mark -h | --help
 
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -108,7 +108,7 @@ By default, mark provides several built-in templates and macros:
 Usage:
   mark [options] [-u <username>] [-p <token>] [-k] [-l <url>] -f <file>
   mark [options] [-u <username>] [-p <password>] [-k] [-b <url>] -f <file>
-  mark [options] [-u <username>] [-p <password>] [-k] [-n] -c <file>
```
kovetskiy, line 111: Nice catch, thanks

pkg/mark/markdown.go
```diff
@@ -90,3 +90,13 @@ func CompileMarkdown(
 
 	return string(html)
 }
+
+// dropH1Markdown will drop leading H1 headings to prevent
+// duplication of or conflict with page titles.
+func DropH1Markdown(
+	markdown []byte,
+) []byte {
+	h1 := regexp.MustCompile(`^#[^#].*\n`)
```
kovetskiy, line 99: The current implementation will drop all h1 headings from the entire document, is it expected? I thought you wanted to remove only the first h1 heading.

lukiffer: I believe the `^` on a non `/m` regexp will only match the beginning of the document rather than the beginning of the individual line. This makes context (probably unnecessarily) important – when passing the whole markdown document from [main.go](https://github.com/kovetskiy/mark/blob/master/main.go#L135), this works as intended. If it were to be used from [meta.go](https://github.com/kovetskiy/mark/blob/master/pkg/mark/meta.go#L41) inside of the `scanner.Scan()` loop, this would have undesirable effects (like stripping comments from bash codeblocks, etc.)

`DropDocumentLeadingH1()` seems a bit verbose, but probably warranted given the potential side effects. I'll also make this clearer in comments. Let me know if you have another suggestion.

kovetskiy:
Merged, thanks for the contribution!

PR #37, opened 2020-11-26
Remove unused variable

jcavar:


kovetskiy:
Nice catch, thanks.

How did you actually find that it's unused in the first place?

jcavar:
It was luck. I downloaded source code and ran `make build`. I got an error that I am not inside a git repository. Then I went to check how git is used and realized that the `BRANCH` variable is not used anywhere.

PR #39, opened 2020-11-29
Incorporate the collapsible code block

snejus:
Something I use in the docs often is the `<details><summary>` html block which allows to embed code blocks of any size without losing out on the readibility - you can open and close the thing as you like.

Now the problem is that Confluence doesn't parse it (or does it in a weird way - I don't remember) and instead they have their own _collapsible_ something, which is enabled through a flag in the `code` macro. This PR enables it, as well as the `title` field.

It's a bit of funky way to do it, but luckily `github`'s markdown/codeblock parser is flexible enough to allow this:

    ```python collapse title Hello this is a Beautiful codeblock

    def func():
        print("Beautiful")

    # ... and 200 more lines of funk
    ```

And on confluence it looks this way:

image

Best part is that github's version looks as usual, seems like they just ignore the extra words.

If we wrap it in

    <details>
    <summary>Hello this is a Beautiful codeblock</summary>

    ```python collapse title Hello this is a Beautiful codeblock

    def func():
        print("Beautiful")

    # ... and 200 more lines of funk
    ```
    </details>

We get

<details>
<summary>Hello this is a Beautiful codeblock</summary>

```python collapse title Hello this is a Beautiful codeblock

def func():
    print("Beautiful")

# ... and 200 more lines of funk
```
</details>

The lines that contain these tags have to be removed before uploading to confluence but that's specific to how I use it so may not apply to everyone - otherwise piping the file through `sed /<\/\?details>\|<\/\?summary>/d` removes them.

Let me know if I should remove the `printfs` from the template - it's a rather subjective change so I'm happy to let them go if you have doubts about them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ must stay in the front _if it is given_:
 
     [<language>] ["collapse"] ["title" <your title>]
 
-[code block macro]: https://confluence.atlassian.com/doc/code-block-macro-139390.html
+[Code Block Macro]: https://confluence.atlassian.com/doc/code-block-macro-139390.html
 
 ## Template & Macros
 
--- a/README.md
+++ b/README.md
@@ -92,6 +92,33 @@ for example:
        Ticket: ${0} -->
 ```
 
+### Code Blocks
+
+If you have long code blocks, you can make them collapsible with the [Code Block Macro]:
+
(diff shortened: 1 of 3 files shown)


kovetskiy:
Thank you!

Issue #17, opened 2020-02-20
Error: "method setContentPermissions taking 3 parameters" when trying to lock page


Netzeband:
Hello

I'm using Confluence 7.2.0 and when I upload a markdown file together with the request to lock the page (option `-k`) I get the following error code:

```
2020-02-20 06:24:30 [FATAL] 'true' response expected, but 'map[error:map[code:-32602 data:<nil> message:Request parameter types did not match method parameter types (method setContentPermissions taking 3 parameters)] id:<nil> jsonrpc:2.0]' encountered
```

Just uploading the page works well. In the `api.go` file I can see, that the function `RestrictPageUpdatesServer` passes indeed 3 arguments to the API call, so I don't understand why Confluence is complaining here. Maybe the API changed?

elgreco247:
The signature of the json-rpc call in `RestrictPageUpdatesServer` wasn't right: #45

kovetskiy:
Closing because fixed by #45 
Thanks to @elgreco247

PR #45, opened 2020-12-07
Fix RestrictPageUpdatesServer (#17)

elgreco247:


kovetskiy:
Hey, thank you for the fix!

GitHub did not recognize the email in your commit so it did not connect it with your GitHub account. This is why the avatar on the commit is grey. Would you like to fix it before the merge?

elgreco247:
Fixed the email in the commit.

PR #62, opened 2021-01-27
Add support to reading password from stdin

emead-indeed:
I don't like passwords in configs or my bash history. This MR checks if the password is - and then reads from stdin instead. Blocks otherwise like other applications which do the same.

This allows you to pull your password out of a password manager and push it into the application with a pipe.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -2,7 +2,9 @@ package main
 
 import (
 	"errors"
+	"io/ioutil"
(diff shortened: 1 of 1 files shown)


kovetskiy:
Thank you!

Snapp949 (commented):
Fuck man

PR #85, opened 2021-04-02
Add support for fileglobs

wbh1:
This change is backwards compatible. The only odd behavior I encountered during testing is that I had to always quote my glob patterns. This seems to be a limitation of `docopt`. For example:
```bash
❯ mark -f "helpful_cmds/*.md"
```

Quoting remains unnecessary if being used as before (one file at a time). 

Also worth noting that Go does not support double star matching yet (https://github.com/golang/go/issues/11862).


kovetskiy:
Thank you!

PR #89, opened 2021-06-17
feat(sidebar): Add sidebar support

beeme1mr:
Added an optional metadata header that allows users to set content in the right-side column. It does explicitly set the `Layout` to `article` when this option is enabled.


kovetskiy:
Thanks for the contribution and your patience :)

Issue #42, opened 2020-12-02
Leading closing html comment in page content


willsmith9182:
Howdy, 

am i missing something? from what i see all are opened and closed correctly but i get a leading `-->` on pages in confluence

.md file:
```
<!-- Space: DD -->
<!-- Parent: On-boarding docs -->
<!-- Parent: Systems Development -->
<!-- Parent: Application -->
<!-- Parent: BuildTools -->
<!-- Title: Repositories using BuildTools -->

<!-- Macro: :autogen:
     Template: ac:status
     Title: AUTOMATED
     Color: Yellow -->

:autogen:
this document is generated, do not edit manually.

----

some other content (no html comments at all in this section)

```

Expected: 
no leading `-->` in content

Actual:
leading `-->` in content of page, can bee see in raw macroRenderedOutput in trace request as `<p>&ndash;&gt;</p>\n\n`

![image](https://user-images.githubusercontent.com/11752440/100912062-ee88a180-34c7-11eb-8c8c-a525bb5181c1.png)

kovetskiy:
Hi, I was not able to reproduce this on 3.5.

willsmith9182:
I deleted and re-created my page, it's still there using 3.5 too. 

I can see in a trace log there's a `-->` but i can't figure out where it's come from 👍 
```
2020-12-06 22:49:09 [TRACE]: rendering markdown:
                             -->

                             <ac:structured-macro ac:name="status"><ac:parameter ac:name="colour">Yellow</ac:parameter><ac:parameter ac:name="title">AUTOMATED</ac:parameter><ac:parameter ac:name="subtle">false</ac:parameter></ac:structured-macro>
                             this document is generated, do not edit manually.

                             ----

                             ## Build Tools Type: Cake
```

willsmith9182:
I narrowed it down, removing the following directives removed the trailing `-->`

```<!-- Parent: On-boarding docs -->
<!-- Parent: Systems Development -->
<!-- Parent: Application -->
<!-- Parent: BuildTools -->```

i wdoner if it's spaces in the parents or just multiples that causes it?

bencurrandev:
I am getting the same issue, the length of the closing html comment differs based on how many parent tags are present. I do also have spaces in the parent names, but there is no correlation between the spaces and tailing characters. They still appear if I remove all spaces from the parents. Exact same as @willsmith9182 in the trace as well. Happened in 4.2 and 5.2.1.

johns10:
[...]

PR #91, opened 2021-07-07
Change offset to 2 characters instead of 1

johns10:


kovetskiy:
Hello, why?

johns10:
This fixes https://github.com/kovetskiy/mark/issues/42. There's detailed comments on that ticket, but I'll summarize here:

I pulled down the repo so I could repro locally and characterize. I made 2 test files, one with only the necessary front matter, and another that duplicated the attachment in the front matter multiple times.

[test1.md](https://github.com/kovetskiy/mark/files/6791331/test1.md)
[test2.md](https://github.com/kovetskiy/mark/files/6791332/test2.md)
124802770-530c5300-df26-11eb-9783-1165abbd077e

test1.md results:
```
>

# Test File
```
test2.md results:
```
tachment: images/1.png -->

# Test File
```
It's not 1:1 for lines of front matter to characters left in the markdown, but it's close.

I instrumented by inserting fmt.Printf("MD: %s", markdown) into main.go at line 165.

I believe the issue is that we must adjust the offset for 2 characters to account for both characters of the line feed.

kovetskiy:
Thanks for the explanation, your patience, and your willingness to contribute to the project.

I've finally got my hands on this ticket. I checked your test1/test markdown files and found why you need `+2` there. This is because you used `\r\n` as new line breaks. It will definitely work for these files but will not work for 99% of other files created on Linux/macOS since on Unix we don't use \r for breaking lines.

More about new line format difference: https://en.wikipedia.org/wiki/Newline?oldformat=true

What I'd suggest is to rollback the current change and make `ExtractData` expect newline break to be `\n` but in the `processFile` in `main.go` drop all the \r\n breaks and replace them with `\n`.

It would be great to have a test for this, but a comment on top of string replacement would work for me too.

kovetskiy:
Changing offset to +2 also can lead to a panic if a file uses \n instead of \r\n.

For example, you can take your `test2.md` and convert it to unix line breaking standard:
```
sed -i $'s/\r$//' test2.md
```

running mark against it will  result in panic:
```
panic: runtime error: slice bounds out of range [1020:1002]

goroutine 1 [running]:
github.com/kovetskiy/mark/pkg/mark.ExtractMeta(0xc000328000, 0x3ea, 0x3eb, 0x3ea, 0x3eb, 0x0, 0x0, 0x3c, 0x10)
```

johns10:
Ahhh, I see!  I'll pick this up on my next loop. I thought it was because there were 2 characters in \n. Honestly I may just fix my file. It's curious it's coming in with those line breaks considering I'm authoring in VSCode/Windows.

kovetskiy:
It looks like you've solved your issue by just using \n instead of \r\n. Feel free to re-open the issue!

PR #77, opened 2021-03-26
Add ac:box macro

rofafor:
README example:
image

PR #93, opened 2021-07-10
update the panel format

Taldrain:
I've tried to use the panel format but the output was not what I was expecting

master

After applying the fix I have the expected output:

fork

---

I don't really know why the feature from https://github.com/kovetskiy/mark/pull/77 was not working. Is it a change from the new Confluence editor?


kovetskiy:
Thanks for the contribution!

It probably is a change from the new confluence editor.

Issue #88, opened 2021-05-14
Updating the root article of a space


rslotte:
Is it possible to update the Overview or the root article of a given Confluence space?

When checking the advanced page info we can see this page does not have any parents. 

So I would have thought we could do this:
```
<!-- Space: TT -->
<!-- Title: Test Team Home -->
# Test Team Home

Welcome!
```

However, when running mark against this file I get the following response:
```
2021-05-14 13:30:00.303 FATAL unable to resolve page location
                              └─ page "Test Team Home" has no parents
```

Anything I'm missing? Thanks

mrueg:
This works now — the scenario in the report publishes successfully against current `master`.

Verified with exactly the shape you posted: a document with `Space` and `Title`, no `Parent` header, where the title matches the space homepage. mark recognises that the homepage legitimately has no ancestors, updates it in place, and leaves it at the root of the space.

```markdown
<!-- Space: TT -->
<!-- Title: Test Team Home -->

Welcome!
```

I could not pin down which change fixed it — the homepage branch in ancestry validation has been there a while, so this was resolved at some point without the issue being updated. There is now a regression test for it (`TestHomepageItselfIsStillPublishable`) so it stays fixed.

One neighbouring case was still broken and is being fixed in #882: a page sitting *at* the space root **beside** the homepage rather than being the homepage. mark places a document with no declared parents under the space root page, so it saw such a page as misplaced and refused with the same `has no parents` wording — which was inescapable, since declaring the parent it ought to have is exactly what produced that state. It is adopted and moved now rather than refused.

If your Overview page is registered as the space homepage, the first case covers you and this already works. If you hit `has no parents` on something that is not the homepage, that is #882.

Closing as fixed — please reopen if you can still reproduce it.

PR #101, opened 2021-08-08
allow space home update and its children pages

eitchugo:
This patch allows editing a Space Home Page as defined in Confluence Space Settings. It also allows to create/update the home page's direct children using a parent meta:

```
<!-- Parent: Space Home -->
```

If a page isn't a direct child of the space homepage, the above parent *should not* be present.

Notice that this will also query confluence API on `/space/{spaceKey}` to check which page is defined as the Space Home. This happens on every run.

Should resolve Issue #88

### DIFF
--- a/pkg/confluence/api.go
+++ b/pkg/confluence/api.go
@@ -474,18 +474,10 @@ func (api *API) UpdatePage(
 	nextPageVersion := page.Version.Number + 1
 	oldAncestors := []map[string]interface{}{}
 
-	if page.Type != "blogpost" {
-		if len(page.Ancestors) == 0 {
-			oldAncestors = nil
-			//return fmt.Errorf(
-			//	"page %q info does not contain any information about parents",
-			//	page.ID,
-			//)
-		} else {
-			// picking only the last one, which is required by confluence
-			oldAncestors = []map[string]interface{}{
-				{"id": page.Ancestors[len(page.Ancestors)-1].Id},
-			}
+	if page.Type != "blogpost" && len(page.Ancestors) > 0 {
+		// picking only the last one, which is required by confluence
+		oldAncestors = []map[string]interface{}{
+			{"id": page.Ancestors[len(page.Ancestors)-1].Id},
 		}
 	}
 
--- a/pkg/confluence/api.go
+++ b/pkg/confluence/api.go
@@ -30,6 +30,18 @@ type API struct {
 	BaseURL string
 }
 
+type SpaceInfo struct {
+	ID   int    `json:"id"`
+	Key  string `json:"key"`
+	Name string `json:"name"`
+
+	Homepage PageInfo `json:"homepage"`
+
+	Links struct {
+		Full string `json:"webui"`
+	} `json:"_links"`
+}
+
 type PageInfo struct {
 	ID    string `json:"id"`
 	Title string `json:"title"`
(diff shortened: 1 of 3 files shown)


pkg/mark/mark.go
```diff
@@ -32,8 +32,25 @@ func ResolvePage(
...
+	homepage, err := api.FindHomePage(meta.Space)
+	if err != nil {
+		return nil, nil, karma.Format(
+			err,
+			"can't obtain home page from space %q",
+			meta.Space,
+		)
+	}
+
+	skip_home_ancestry := false
```
kovetskiy, line 45: Could you convert skip_home_ancestry and is_homepage to camelCase? In Go we don't really use snake_case at all :)

pkg/confluence/api.go
```diff
@@ -444,15 +476,16 @@ func (api *API) UpdatePage(
...
-			return fmt.Errorf(
-				"page %q info does not contain any information about parents",
-				page.ID,
-			)
-		}
-
-		// picking only the last one, which is required by confluence
-		oldAncestors = []map[string]interface{}{
-			{"id": page.Ancestors[len(page.Ancestors)-1].Id},
+			oldAncestors = nil
```
kovetskiy, line 479: No need to set it to nil here since you can just make an opposition of the if-condition, like this:

```go
if page.Type != "blogpost" && len(page.Ancestors) > 0 {
	oldAncestors = []map[string]interface{}{
		{"id": page.Ancestors[len(page.Ancestors)-1].Id},
	}
}
```

pkg/mark/mark.go
```diff
@@ -32,8 +32,25 @@ func ResolvePage(
...
+	if len(meta.Parents) > 0 {
+		if homepage.Title == meta.Parents[0] {
+			skip_home_ancestry = true
+		}
+	}
+
 	ancestry := meta.Parents
-	if page != nil {
+	if page != nil && skip_home_ancestry == false {
```
kovetskiy, line 53: `&& !skipHomeAncestry`

kovetskiy (changes requested):
Overall it looks good to me, but we still need to fix a few code style problems. Thank you for your contribution!

kovetskiy:
You can also see that you've committed with (probably) your work email, do you want to force push your commit with your GitHub's email? GitHub didn't recognize your email and didn't attach your commit to your profile.

https://github.com/kovetskiy/mark/commit/6aa37507a3e79ce5cc5901f7414b4f5f8b5e7381.patch

pkg/mark/mark.go
eitchugo: Oh, thanks for telling me :) I'm from the Python world lol

eitchugo:
@kovetskiy that's OK, I don't mind not having my github email on the commits, have it your way :D

Issues should have been resolved, let me know if there's anything else I could do

pkg/mark/ancestry.go
```diff
@@ -104,11 +104,26 @@ func ValidateAncestry(
...
+		if page.ID == homepage.ID {
+			log.Debugf(nil, "page is homepage for space %q", space)
+			isHomepage = true
+		} else {
+			return nil, fmt.Errorf(`page %q has no parents`, page.Title)
+		}
 	}
 
-	if len(page.Ancestors) < len(ancestry) {
+	if len(page.Ancestors) < len(ancestry) && isHomepage != true {
```
kovetskiy, line 126:
```suggestion
	if !isHomepage && len(page.Ancestors) < len(ancestry) {
```

pkg/mark/mark.go
```diff
@@ -32,8 +32,25 @@ func ResolvePage(
...
+	skipHomeAncestry := false
+	if len(meta.Parents) > 0 {
+		if homepage.Title == meta.Parents[0] {
+			skipHomeAncestry = true
+		}
+	}
+
 	ancestry := meta.Parents
-	if page != nil {
+	if page != nil && skipHomeAncestry == false {
```
kovetskiy, line 53:
```suggestion
	if page != nil && !skipHomeAncestry {
```

kovetskiy:
Thanks!

Issue #94, opened 2021-07-14
Sync failure on image


mantzas:
Hi,

when creating a page with image attachment the process works perfectly fine.
After updating the image attachment and running sync again we face a lot of times failures with the following output:
`2021-07-09 16:54:30.450 FATAL Confluence API returned unexpected status: 500 , output: ""`
The workaround is to delete the attachment in Confluence and sync again.
Is this something other face also?
Thanks

kovetskiy:
Hi, I haven't seen this behavior before, could you provide steps to reproduce it?

mantzas:
The steps are pretty simple and that's why I am puzzled by it.
- Create a Markdown file with an image attachment
- On the first page creation everything works well
- Changing the image (if the image does not change no problem exists)
- Sync again with mark which results in an error

The command used in both sync is (obviously I have changed the arguments):

`mark -u username -p password -b https://confluence.company.com -f README.md`

and when I sync the second time, with and existing image attachment, it fails with:

```
2021-07-14 12:58:42.057 INFO  processing README.md
2021-07-14 12:58:45.384 INFO  page will be stored under path: Page1 > Page2 > Page3 > Page4
2021-07-14 12:58:45.479 INFO  updating attachment: "doc/architecture/image.png"
2021-07-14 12:58:46.447 FATAL unable to create/update attachments
                              └─ unable to update attachment "doc/architecture/image.png"
                                 └─ Confluence REST API for creating attachments returned 0 json objects, expected at least 1

```

this is the header of the markdown:
```
<!-- Space: DT -->
<!-- Title: Page1 -->
<!-- Parent: Page2 -->
<!-- Parent: Page3 -->
<!-- Attachment: doc/architecture/image.png -->
```

and this is how the image is embedded in the markdown:

```
<img alt="image" src="doc/architecture/image.png" width="1187" />
```
Hope this helps

mantzas:
BTW this happens even if i use the markdown way of inlining images

```
![Image](doc/architecture/image.png)
```

kovetskiy:
You are using a hosted version of confluence right? Which version you are on?

mantzas:
`Powered by Atlassian Confluence 6.15.2`

rafadc:
I am having the same problem. I run an example response with `--trace`. Could it be related to versioning of the attachments and the payload being different? When you run it the second time it looks to me like you are trying to create a second version instead of creating a completely new attachment. 


```
{
   "id":"att3037331492",
[...]

PR #110, opened 2021-08-25
Response for updated attachment corrected for jira cloud

rafadc:
Fixes issue #94

I am debugging this and looks like for `POST /wiki/rest/api/content/{id}/child/attachment/{attachmentId}/data` the API returns a 

```
{ attachment }
```

instead of

```
{ "results": [attachment] }
```

The [docs on the cloud version](https://developer.atlassian.com/cloud/confluence/rest/api-group-content---attachments/#api-wiki-rest-api-content-id-child-attachment-attachmentid-data-post) do not specify the return format so I cannot be sure the hosted version works the same way exactly.

Be aware that I only have a Jira Cloud available so maybe this breaks Jira hosted.

### DIFF
--- a/pkg/confluence/api.go
+++ b/pkg/confluence/api.go
@@ -262,12 +262,7 @@ func (api *API) UpdateAttachment(
 		return AttachmentInfo{}, err
 	}
 
-	var result struct {
-		Links struct {
-			Context string `json:"context"`
-		} `json:"_links"`
-		Results []AttachmentInfo `json:"results"`
-	}
+	var result AttachmentInfo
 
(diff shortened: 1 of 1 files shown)


kovetskiy:
Yikes, the problem with this PR is that I'm 99% sure that this code worked before which means if we merge this, the program will not work on older versions of Confluence, right? I guess we need to check the version of Confluence or somehow try both formats.

kovetskiy:
Closing since the problem was resolved in #9cde833a181303bb851247ae263a554edf65113c

PR #133, opened 2021-11-05
Children display macro

jevfok:
Added template for [Children Display macro](https://support.atlassian.com/confluence-cloud/docs/insert-the-children-display-macro/).


kovetskiy:
Thank you for the contribution!

PR #138, opened 2021-11-18
Change log level when files are not found

mmiranda:
For CI purposes, when files are not found the process should continue without any hassle.

### DIFF
--- a/main.go
+++ b/main.go
@@ -121,7 +121,7 @@ func main() {
 		log.Fatal(err)
 	}
 	if len(files) == 0 {
-		log.Fatal("No files matched")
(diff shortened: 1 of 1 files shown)


kovetskiy:
Hi, thanks for the contribution and your patience with this PR. 

I understand the needs behind it, but I'm afraid I can't it merge just like that because there are people expecting mark to fail if the given files not found. Can we add an optional flag to introduce this behavior that you want?

mmiranda:
Thank you for the feedback @kovetskiy, I will add a new flag for it and keep the old behavior as default

kovetskiy:
Thanks!

PR #145, opened 2021-12-12
add inline comment support

guoweis-outreach:


guoweis-outreach:
@kovetskiy how did the `'` become `&lsquo;`? the escape I did doesn't handle that well.

kovetskiy:
Hi, I've checked the draft version, the problem with `&lsquo;` comes from the fact that Confluence replaces ' with ’ on the server side when we push HTML via API.

kovetskiy:
Also found a problem:

Original markdown:
```
aaa **bold** bbb
```

Then select the entire paragraph and comment, OriginalSelection of an inline comment structure will be `aaa bold bbb` so MergeComments() func will not be able to replace the content since in the original content the phrase `bold` was wrapped up in two asterisks.

Actually have no idea how to come up with string replacement in such scenarios which will be quite frequent I suppose.

guoweis-outreach:
good catch. I tried the case you mentioned. In MergeComments, on the input, I got `aaa <strong>bold</strong> bbb`, what if we just remove the <strong> and probably with some space handling, this might work?

guoweis-outreach:
i'll create some test cases for these.

kovetskiy:
> good catch. I tried the case you mentioned. In MergeComments, on the input, I got `aaa \<strong\>bold\</strong\> bbb`, what if we just remove the `\<strong\>` and probably with some space handling, this might work?

That's strange since I've got OriginalSelection to be `aaa bold bbb` without <b> or <strong> tags. I'm surprised that they don't pinpoint at a specific location like line/column or at least the number of byte where the comment starts.

I'm not sure if string replacement is really a good way to handle the issue.

guoweis-outreach:
the alternative I thought about was xml path, however, since everything starts with <p>, it wasn't unique enough... Not sure what else can be done other than string replacement here, but I'm open to any other approach. I'll keep going with this to see if the end result is satisfactory.

guoweis-outreach:
the alternative I thought about was xml path, however, since everything starts with <p>, it wasn't unique enough... Not sure what else can be done other than string replacement here, but I'm open to any other approach. I'll keep going with this to see if the end result is satisfactory.

guoweis-outreach:
> That's strange since I've got OriginalSelection to be `aaa bold bbb` without `<b>` or `<strong>` tags

i also got OriginalSelection as `aaa bold bbb` from the inlineComment. My mentioning of `aaa <strong>bold</strong> bbb` is the body when it comes to MergeComment

guoweis-outreach:
the other approach is to download the original body.storage, and compare with the current one to inject the marker, but that is much harder, as it's diffing 2 xml documents...

guoweis-outreach:
@kovetskiy i think about this more. yes, there is no way to fix this ONLY by string comparison. Fundamentally, there is no way we could solve this accurately as the user's edition is a lossy info to us. I think there should be 2 approaches together

1. we could have some heuristics to determine which comment should go where
2. for really motivated users, i wonder if we should invent a syntax so they could specify the comment in the exact location of their doc? Yes, this will be complicated as they will need to fetch the refID, etc... 

Thoughts?

kovetskiy:
@guoweis-outreach I didn't really come up to any ideas that wouldn't require **a lot** of work.

Maybe we could just lock page for comments so people would need to comment the entire page in the comments section instead?

guoweis-outreach:
@kovetskiy i think this one is important. I used mark to sync, but whenever someone comments on the page, I get afraid to update or even edit the page as I'm worry that I'll lose it. Turning off the ability to comment inline can work but it causes lot of inconvenience.

I wonder as the first step, does it make sense to create a way to sync the comment from confluence to the markdown? ie. If we could create a syntax to mark the inline comment in the markdown file, and safely preserve it, then at least people won't feel worry about losing valuable comments.

Thoughts?

Skeeve:
> I wonder as the first step, does it make sense to create a way to sync the comment from confluence to the markdown? ie. If we could create a syntax to mark the inline comment in the markdown file, and safely preserve it, then at least people won't feel worry about losing valuable comments.
> […]
> Thoughts?

Tbh: I'm not a hard-core confluence user and never will be. So maybe I miss the point of inline comments. Maybe it would be best to not allow a push when comments are in the page `mark` wants to update.

Maybe the comments have to be integrated somehow into the markdown and removed from the page first? I'm thinking here about comments which point to mistakes in the page or missing stuff. But as said: Maybe I have to less a clue of how confluence is intended to be used.

andymac4182:
I have started on this similar problem in my project. Might provide some inspiration for you when looking at how to find where to put comments. https://github.com/markdown-confluence/markdown-confluence/blob/main/packages/lib/src/AdfProcessing.ts#L285-L500

I am using https://en.wikipedia.org/wiki/Levenshtein_distance calculation on the text around the comment to find the "best" place to put it back. Need to add in the same for when you modify the comment text.

mrueg:
Thanks for the PR, I've reimplemented your suggestion in #768

Issue #31, opened 2020-10-21
Resolving relative images in subdirectories


Voronenko:
For markdown documents in subdirectories referring relative images,like

```markdown

![alt](diagrams/context/context.svg)

```

tool  is able to find image only in scenario, when it executed from the same directory where file resides.

If I will try to execute it passing file path, like 
```sh
mark -u $CF_USER -p $CF_TOKEN -b $CF_DOMAIN -f docs/subdirectory/$file
```

it will file.

So far workaround is manual traversing of dirs one by one, but would be nice if relative references could be resolved.

```sh
declare -a dirs=("docs", "docs/subfolder1", "docs/subfolder2" )

for d in ${dirs[@]}; do

(
cd $d
for file in $(find -maxdepth 1 -type f -name '*.md'); do
  echo "> Sync $file";
#  echo mark -u $CF_USER -p $CF_TOKEN -b $CF_DOMAIN -f $file
  mark -u $CF_USER -p $CF_TOKEN -b $CF_DOMAIN -f $file || exit 1;
  echo;
done
)

done
```

lukiffer:
The workaround listed does work. I've modified it a bit to make it a little more flexible until I've got time to look for a fix in go:

```bash
set -e
while read -r -d '' file; do
  echo "Syncing $file..."
  pushd "$(dirname "$file")" > /dev/null
    filename="$(basename "$file")"
    mark -k -f "$filename"
  popd > /dev/null
done < <(find . -type f -name "*.md" -print0)
```

GrantAdkins:
> The workaround listed does work. I've modified it a bit to make it a little more flexible until I've got time to look for a fix in go:
> 
> ```shell
> set -e
> while read -r -d '' file; do
>   echo "Syncing $file..."
>   pushd "$(dirname "$file")" > /dev/null
>     filename="$(basename "$file")"
>     mark -k -f "$filename"
>   popd > /dev/null
> done < <(find . -type f -name "*.md" -print0)
> ```

Unfortunately this workaround fails when using it in a Bitbucket pipeline yaml file, resulting error is "syntax error near unexpected token `do'"

Without this workaround, I can get the files attached in Confluence by specifying the full path in the  meta tags but they're not embedded in the page itself as the path using in the markdown doesn't exactly match the full path.

chipbite:
Would be really good to have this support in place, right now, having all image assets in the same folder quickly gets problematic for us.

kovetskiy:
[...]

PR #151, opened 2022-01-06
Fix for subdirectories related attachments (issue #31)

klysunkin:
Fixed by setting the base path to the current file directory. Also replaced map with an array for 'replacements' as map has the same key and value.

### DIFF
--- a/pkg/mark/attachment.go
+++ b/pkg/mark/attachment.go
@@ -35,17 +35,14 @@ func ResolveAttachments(
 	api *confluence.API,
 	page *confluence.PageInfo,
 	base string,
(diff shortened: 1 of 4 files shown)


pkg/mark/attachment.go
```diff
@@ -160,6 +155,22 @@ func ResolveAttachments(
 	return attaches, nil
 }
 
+func prepareAttachments(base string, replacements []string) ([]Attachment, error) {
```
kovetskiy, line 158: doesn't look like we need this `error` in the return variables, but it's fine tho

kovetskiy:
Thanks!

Issue #95, opened 2021-07-14
Support page title from first h1


bentinata:
Additionally, this can be used to drop `--drop-h1` options, since usually `h1` is used as title anyway.

mrueg:
I believe this is resolved. Please let me know if it's not.

Issue #105, opened 2021-08-10
Support double star (**) glob syntax


SleepWalker:
Hello,

it turns out that default golang Glob function does not support common globstar (`**`) syntax. This syntax allows matching of files in subdirs too.

Here are some useful links:

- https://stackoverflow.com/questions/26809484/how-to-use-double-star-glob-in-go
- https://github.com/bmatcuk/doublestar

Harvzor:
I've got this file structure:

```
root@903a83c00b08:/app/docs# tree
.
|-- docs
     |-- foo.md
     |-- bar.md
     |-- baz
         |-- qux.md
```

I ran with the file globbing pattern of `-f "docs/**/*.md"` in the base folder and it only found `qux.md`.

So it seems that this tool does support the `**` syntax but that it won't find files in the `docs` folder, but only subdirs of of `docs`, but probably not subdirds of those subdirs 😄

oradwell:
I've created #392 to fix this. anyone still interested in it?

PR #153, opened 2022-01-17
Space from config and title from h1

klysunkin:
This PR resolves:
Support defining space from config #96
Support page title from first h1 #95

--space and --title-from-h1 options were added. If Space is defined both in metadata and in a command line then the value from file metadata overrides the value from CL. If Title is defined in metadata it's being used regardless of is H1 header was set in a file or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -438,7 +438,9 @@ mark -h | --help
 - `-c <path>` or `--config <path>` — Specify a path to the configuration file.
 - `-k` — Lock page editing to current user only to prevent accidental
     manual edits over Confluence Web UI.
+- `--space <space>` - Use specified space key. If not specified space ley must be set in a page metadata.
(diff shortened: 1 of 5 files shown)


kovetskiy:
Hey, this looks good to me and I'm ready to merge and release the new version. I'm just wondering, what is the usecase where you need such a behavior as specifying Space through flags?

klysunkin:
From my perspective Space as well as a base URL does not relate to a page but rather set up a context. So, I believe it'll be useful for those who don't want to couple all pages with a space key. One use case from my mind is to have documentation for release and for WIP in different branches and publish to different spaces.
Also, with https://github.com/kovetskiy/mark/pull/151 merged and https://github.com/kovetskiy/mark/issues/105 fixed it will be possible to publish pages hierarchy with one command.

kovetskiy:
Thanks!

PR #154, opened 2022-01-17
feat: Add support of confluence server.

vpommier:
Add support of [personal access token authentication](https://confluence.atlassian.com/enterprise/using-personal-access-tokens-1026032365.html) for confluence server.

### DIFF
--- a/main.go
+++ b/main.go
@@ -39,7 +39,7 @@ type Flags struct {
 }
 
 const (
-	version = "7.0"
+	version = "7.1"
 	usage   = `mark - a tool for updating Atlassian Confluence pages from markdown.
 
 Docs: https://github.com/kovetskiy/mark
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -72,6 +73,7 @@ Options:
   -c --config <path>   Use the specified configuration file.
                         [default: $HOME/.config/mark]
   --ci                 Runs on CI mode. It won't fail if files are not found.
+	--confluence-cloud   Use Confluence Cloud or Confluence hosted server authentication method.
```
kovetskiy, line 76: Something wrong with indentation here

kovetskiy (changes requested):
I'm not sure it's a good way to break everyone's workflow and make them use some new flag which was not required before.

Can we just always send the Authorization token for both cloud and server?

vpommier:
> I'm not sure it's a good way to break everyone's workflow and make them use some new flag which was not required before.
> 
> Can we just always send the Authorization token for both cloud and server?

When i send the Authorization token it works only for server but not cloud.
And even if it worked there is no guarantee that this behavior (having basic auth + bearer token at the same time) will be supported in the future. It is the best way to attract issues in my opinion.

The thing that i can do is to require the new flag only for confluence server, in that way the new version will doesn't break the actual workflow.

I made these kind of changes to rapidly test if the tool can solve my issue with this authentication method. But i think the proper way to fix the code is to patch the code this way:
* [Headers should be copied instead of passing the reference of them `r.Headers.Clone()`. Because it create side effet.](https://github.com/kovetskiy/gopencils/blob/master/resource.go#L54)
* [Provide something more than just basic auth.](https://github.com/kovetskiy/gopencils/blob/master/api.go#L50)

main.go
vpommier: Fixed

vpommier:
Hi @kovetskiy i pushed modifications any news please ?

kovetskiy:
Works great, thanks!

PR #128, opened 2021-10-25
feat: support mermaid graphs

carnei-ro:

PR #167, opened 2022-03-14
Add mermaid support

dreampuf:
Hi, me again. Thanks for maintaining the cool tool. I'm asking to add a cool feature.

Before [Github officially enabled mermaid support](https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/), I'm a heavy user of markdown. I think we all know the best idea of markdown is to provide the ability to integrate the git workflow. 
The diagrams are the same. That's the mermaid comes from. The basic idea isn't new. Graphviz and other tools can do the same thing. But they can't either be easily integrated with or provide the fancy built-in diagram types.

There are two ways to make mermaid integrated with confluence. Client-side render and server-side render. My approach is kinda server side. It renders the code if it found them before you upload your page. And render the picture as an attachment. Upload it and insert the picture in a proper place (before the code block and set the block as collapse).

Talking to the cons, the image size, and third-party dependencies. The image size is from 30MB to 389MB. Mermaid render [needs](https://github.com/mermaid-js/mermaid/issues/559) a browser environment. It makes mark need to coexist with a browser. But on the other hand, you still need to view your confluence page. This PR comes with chrome-headless.

All the best,
Dreampuf


dreampuf:
image

main.go
```diff
@@ -322,11 +323,19 @@ func processFile(
 		target = page
 	}
 
+	mermaidImages, err := mark.ExtractMermaidImage(markdown)
+	if err != nil {
+		log.Fatal(err, "unable to render mermiad")
```
lukeamaral, line 328: Typo

DemonWareXT:
I love this idea, but I think it never got pushed right?

dreampuf:
@DemonWareXT Hi, I think there is a way to identify whether a cloud or server is by the domain. According to https://jira.atlassian.com/browse/CLOUD-6999
This is achievable. Let me know if you are interested.

DemonWareXT:
@dreampuf There is a cheap alternative way I think, at least it works with confluence server, I can't check with confluence cloud.
I just set ```language = ""```
Dirty hack I guess, but it could be much easier

dreampuf:
I notice the main branch adopt mermaid support via #128 . The implementation relies on [cloudscript.io](https://marketplace.atlassian.com/apps/1219878/cloudscript-io-mermaid-addon) plugin, which is not free. Our instance doesn't have it. I just set the language to `plaintext` (it could be any from the list of `code` macro [supported languages](https://confluence.atlassian.com/doc/code-block-macro-139390.html))

WildDogOne:
@dreampuf yeah exactly, what is why I will keep running your branch because it works for me that way, I don't want to get some strange paid plugin for that

mrueg:
@dreampuf would you be interested in reviving this MR?

jd-brennan:
> @dreampuf There is a cheap alternative way I think, at least it works with confluence server, I can't check with confluence cloud.
I just set language = ""

Doesn't work for me with confluence cloud. ;-(. I still get the mermaid code displayed below the diagram.

mrueg:
Thanks, I've included your contribution as https://github.com/kovetskiy/mark/pull/288

PR #174, opened 2022-04-12
bugfix: Add delay between create and update calls

### LINKED ISSUE #139: Syncing New Pages Can Error with "409 Conflict"
Possibly a race condition when creating new pages. There seems to be two distinct operations, as indicated by the page's history:

1. Create New Blank Page
2. Sync Page Contents

We are seeing the following error when creating new pages:

```
FATAL Confluence API returned unexpected status: 409 Conflict, output: ""
```

It is believed the above two phases happening in quick succession might be the cause. A resync will allow the second operation to succeed.

JAndritsch:
This helps mitigate a 409 conflict response from Confluence when
creating a new page and then updating it right away. This may not be the
cleanest solution, but it seems to do the trick.

Resolves #139

### DIFF
--- a/main.go
+++ b/main.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"time"
 
 	"github.com/docopt/docopt-go"
 	"github.com/kovetskiy/lorg"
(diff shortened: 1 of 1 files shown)


kovetskiy:
This doesn't seem good to do such a sleep operation, but if it works it works. Thanks!

JAndritsch:
Maybe a better alternative is to add some built-in retry logic whenever 409 conflicts occur. I could explore that option.

PR #175, opened 2022-04-13
bugfix: Force pages to default width.

### LINKED ISSUE #173: Pages are created with inconsistent width settings
I noticed that pages are sometimes being created either at full-width or fixed-width, randomly. I ran `mark` against a folder that contained 12 Markdown files, all with the same content with the exception of the page title metadata. Some of those pages ended up being created either fixed width while some ended up as full width.

Example fixed-width page:

Screen Shot 2022-04-12 at 6 00 39 PM

Example full-width page:

Screen Shot 2022-04-12 at 6 00 54 PM

JAndritsch:
This change is to compensate for an API bug in Confluence where newly
create pages end up getting a random width of either "fixed" or "full".

Fixes #173

### DIFF
--- a/pkg/confluence/api.go
+++ b/pkg/confluence/api.go
@@ -552,6 +552,16 @@ func (api *API) UpdatePage(
 		},
 		"metadata": map[string]interface{}{
 			"labels": labels,
+			// The Confluence API randomly sets page width for some reason:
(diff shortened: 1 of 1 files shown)


kovetskiy:
Thanks!

Issue #73, opened 2021-03-13
Collapsible code sections


antonalekseev:
I wonder what's the reason behind placing Confluence `Code block` with `collapsible = true` inside `Expand` block?

This way the following code:

<pre>
```bash collapse
#!/bin/sh

set -e

exit 0
```
</pre>

translates to

<img width="929" alt="Screenshot 2021-03-13 at 19 58 57" src="https://user-images.githubusercontent.com/2895684/111037934-3fd8f300-841e-11eb-92cf-73b3a87d6f73.png">

and ultimately renders to

<img width="637" alt="Screenshot 2021-03-13 at 20 06 06" src="https://user-images.githubusercontent.com/2895684/111037994-82023480-841e-11eb-99d9-c874455ab31b.png">

In order for a user to see contents of the block they should click `Click here...`, then click `Expand source`, then (in my case, YMMV) whole section collapses again, and when you click `Click here...` again the block is ultimately expanded and the code is visible.

Shouldn't just `{code ... | collapsible = true}` not wrapped with `{expand ...}` suffice?

antonalekseev:
Ah, I see.

@tyler-copilot, with what version of Confluence you observed behaviour you mentioned in #58? I'm on `6.4.2` and `{code ... | collapsible = true}` works as expected for me.

PhilippReinke:
I agree. In newer confluence versions (for me it's 7.13.2) the changes introduced in #58 produce two collapsible boxes. Please check out my PR #179 that reverts it. Judging by what [antonalekseev](https://github.com/antonalekseev) wrote the additional expand macro is not needed for versions >= `6.4.2` (maybe even older) which is already 5 years old. Are you sure you want to support versions that old?

mrueg:
Looks like this has been resolved.

PR #179, opened 2022-04-21
fix: remove double collapse for codeboxes

PhilippReinke:
This fixes a bug where if you used the code macro the renderer would create
an additional expand macro if the collapse option was set. This led to
the code section being collapsible twice.

It is unclear why it was there but maybe there was a reason in earlier confluence versions
that is no longer valid?


PhilippReinke:
This relates to issue #73.

kovetskiy:
Thanks!

PR #183, opened 2022-05-01
Added a command line option that reduces attachment filenames to thei…

Optimierungswerfer:
…r basename only instead of containing the whole local path as a prefix, which can be ugly especially if the local path uses upwards directory movement resulting in something like ".._.._.._dir_Attachment.txt" instead of "Attachment.txt". The command line option is opt-in, which means this addition does not break anything.


Optimierungswerfer:
I messed up this pull request, please excuse this. A clean pull request for this change can be found at #185.

PR #185, opened 2022-05-09
Added a command line option that reduces attachment filenames to thei…

Optimierungswerfer:
…r basename only instead of containing the whole local path as a prefix, which can be ugly especially if the local path uses upwards directory movement resulting in something like ".._.._.._dir_Attachment.txt" instead of "Attachment.txt". The command line option is opt-in, which means this addition does not break anything.

### DIFF
--- a/README.md
+++ b/README.md
@@ -444,6 +444,7 @@ mark -h | --help
 - `--dry-run` — Show resulting HTML and don't update Confluence page content.
 - `--minor-edit` — Don't send notifications while updating Confluence page.
 - `--trace` — Enable trace logs.
+- `--attachment-basename` — Don't include attachment paths in attachment names.
 - `-v | --version` — Show version.
 - `-h | --help` — Show help screen and call 911.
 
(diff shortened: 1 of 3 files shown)


kovetskiy:
Hi! What is your motivation behind making this behavior optional? It feels like we've got a bunch of command line arguments already.

Skeeve:
Why not make it also a configurable option in the `mark` file like I did with `--drop-h1` and `--title-from-h1`.

That way one can make it a default by means of the configuration file.

And maybe instead of having it be a boolean you could make it an integer:

`shorten_filenames=0`=> Do not shorten
`shorten_filename=1` => keep basename
`shorten_filename=<number>` => keep <number> path elements from the end. So 2 would be the parent directory of the attachment plus the filename.

Skeeve:
For just one commandline parameter there are quite a lot of changes.

And what about my suggestion of making it an integer giving the number of elements you want to keep from the filename?

Issue #92, opened 2021-07-09
Handle semantic line breaks when converting markdown file to an HTML file


alexey-tereshenkov-oxb:
Thanks for the work on `mark,` this is an amazing tool! It works great for my needs except for one subtle detail.
I believe it's the "opposite" of another issue: https://github.com/kovetskiy/mark/issues/61

Having this markdown snippet which follows the [semantic line breaks](https://sembr.org/) guide:

```markdown
If you want to save the file,
you can click Save.
```

Will generate this HTML file (which I see by using the `--dry-run` flag):

```html
<p>If you want to save the file,
you can click Save.</p>
```

When this HTML gets rendered in Confluence, the line breaks within the `<p>` tag are respected resulting in awkward breaks:

![image](https://user-images.githubusercontent.com/59651540/125070575-110f1880-e0b0-11eb-82f6-eaf4b48a2b61.png)

I believe most Markdown renderers would ignore single line breaks, so wrapping the text around some length shouldn't affect the rendered outcome. However, I assume Confluence handles the text within the `<p>` tag in a particular way treating those "soft" line breaks as break lines. 

Do you have any control over the HTML generation in the `mark` code to be able to handle it differently to not let Jira make those line breaks?

Many thanks.

kovetskiy:
Hi,
> Do you have any control over the HTML generation in the mark code to be able to handle it differently to not let Jira make those line breaks?
Yes we do have control over the HTML generation and we even have tests that assert that your 1st snippet converts to your 2nd snippet. The problem is that Confluence doesn't care about the fact that newlines in <p> should not be treated as a newline by any RFC.

Do you have any particular ideas on how we can solve this?

alexey-tereshenkov-oxb:
Hi Egor, thanks for looking into this. I've explored a little bit about Confluence handling HTML input and couldn't come up with any bright idea how to work around this.

We have a choice to either let the clients of mark to write and run a pre-processor on their Markdown files to replace single line breaks with an empty space:

```
This is one paragraph
that spans multiple lines
and is sweet and short.

This is a new paragraph
that starts on a next line
and again spanning multiple lines.
```

so that the string above would be converted using something like this (Python snippet): 

```
print(re.sub(r'(?<!\n)\n(?!\n)', ' ', s))
```

to 

```
This is one paragraph that spans multiple lines and is sweet and short.

This is a new paragraph that starts on a next line and again spanning multiple lines.
```

[...]

PR #186, opened 2022-05-18
Add option to remove linebreaks within pagragraph tags

GusAntoniassi:
This merge request addresses #92 by adding the flag `--drop-single-linebreak` which when enabled will strip all `\n` characters within `<p></p>` tags.

Suggestions or changes to the flag name are welcome.

### DIFF
--- a/pkg/mark/markdown_test.go
+++ b/pkg/mark/markdown_test.go
@@ -61,3 +61,18 @@ func TestExtractDocumentLeadingH1(t *testing.T) {
 
 	assert.Equal(t, "a", actual)
 }
+
+func TestRemoveNewlinesWithinParagraphTags(t *testing.T) {
+	html := `<p>This newline
+	should be ignored</p>
+	<span>And this one
+	shouldn't</span>`
+
+	expected := `<p>This newline should be ignored</p>
+	<span>And this one
+	shouldn't</span>`
+
+	actual := RemoveNewlinesWithinParagraphTags(html)
+
+	assert.Equal(t, expected, actual)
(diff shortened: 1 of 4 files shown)


pkg/mark/markdown_test.go
```diff
@@ -61,3 +61,18 @@ func TestExtractDocumentLeadingH1(t *testing.T) {
...
 }
+
+func TestRemoveNewlinesWithinParagraphTags(t *testing.T) {
+	html := `<p>This newline
+	should be ignored</p>
+	<span>And this one
+	shouldn't</span>`
+
+	expected := `<p>This newline should be ignored</p>
+	<span>And this one
```
kovetskiy, line 72: Is this actually what is expected to be a result? Should not it be like `this newline\tshould be ignored`? As far as I see, the main idea was to remove only linebreaks.

nyarly: As far as I'm aware, the goal to to avoid weird breaks in the confluence behavior; that involves flowing text in `<p>`, so `\t` should be removed as well

bennylu2: We'd like this feature as well but I agree that it should only replace linebreaks and not all space. The regex is incorrect.

What's the status of this? It hasn't had updates in a while so looks stale. Wasn't sure if I should put up another PR

nyarly:
I'd really like to see this feature land. @GusAntoniassi are you still working on this?

nyarly:
I think this can be closed in favor of #351 @mrueg

PR #191, opened 2022-05-31
allow templates and macros to include blanks in path

Skeeve:
This small change will allow you to use template files containing blanks.


kovetskiy:
Thanks!

PR #192, opened 2022-05-31
Feature/docker4build

Skeeve:
Just a `docker-compose.yaml` to help developers to compile without installing go.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,13 @@
+version: "3.5"
+services:
+
+  markbuilder:
+    build:
+      context: .
+      dockerfile: Builderfile
+    image: markbuilder:latest
+    volumes:
+    - type: bind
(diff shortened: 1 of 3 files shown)


docker-compose.yaml
```diff
@@ -0,0 +1,37 @@
+version: "3.5"
+services:
+
+  markbuilder:
+    image: golang:latest
+    environment:
+      # MacOS 64-bit
+      - GOOS=darwin
```
kovetskiy, line 8: Do we really need this as default behavior? Can't we just comment all the entries?

Skeeve: Whatever you prefer. These are my settings and it's not an issue at all to change them.

Skeeve: I just changed it so that the values are taken from the environment.

PR #193, opened 2022-05-31
Do not require password and url when compileonly

Skeeve:
When trying to use `--compile-only`, mark complained about missing password and URL. So this PR will set a password of "none" and a URL of "http://localhost" if they are not given but `--compile-only` was requested.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -36,10 +36,13 @@ func GetCredentials(
 	if password == "" {
 		password = config.Password
 		if password == "" {
-			return nil, errors.New(
-				"Confluence password should be specified using -p " +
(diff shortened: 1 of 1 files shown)


kovetskiy:
Is this actually correct thing to do when there is --compile-only flag specified? It feels like the program should not check password/url when it is running in "compile only" mode.

Skeeve:
When in `--compile-only` it currently does check and I didn't want to mess too much with the code. So in case no url or password is given, and only then, I set the defaults.

kovetskiy:
Thanks

PR #196, opened 2022-06-01
Feature/h1 title

Skeeve:
This extends #153 in that there is a new config setting `h1_title` of type bool which will be used when neither a meta Title is set nor `--title-from-h1`  was given.

Same goes for `--drop-h1` which can be set with the bool setting `h1_drop`.

Additionally I added dropping of `h1` when `--compile-only` as this was not done yet.


kovetskiy:
Thanks!

PR #198, opened 2022-06-01
introduce inline templates for Macros

Skeeve:
Templates in macros have to be either predefined ones (`ac:…`) or be defined in additional files.

With this PR one can use inline templates.

### DIFF
--- a/pkg/mark/macro/macro.go
+++ b/pkg/mark/macro/macro.go
@@ -134,12 +134,35 @@ func ExtractMacros(
 				macro Macro
 			)
 
-			macro.Template, err = includes.LoadTemplate(base, template, templates)
-			if err != nil {
-				err = karma.Format(err, "unable to load template")
+			if strings.HasPrefix(template, "#") {
+				cfg := map[string]interface{}{}
 
-				return nil
+				err = yaml.Unmarshal([]byte(config), &cfg)
+				if err != nil {
+					err = karma.Format(
+						err,
+						"unable to unmarshal macros config template",
+					)
+					return nil
+				}
+				body, _ := cfg[template[1:]].(string)
+				macro.Template, err = templates.New(template).Parse(body)
+				if err != nil {
+					err = karma.Format(
+						err,
+						"unable to parse template",
+					)
+					return nil
+				}
+			} else {
+				macro.Template, err = includes.LoadTemplate(base, template, templates)
+				if err != nil {
+					err = karma.Format(err, "unable to load template")
+	
+					return nil
+				}
 			}
+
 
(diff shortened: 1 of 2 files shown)


Skeeve:
Not a bug, but me not understanding the levels of parsing.

pkg/mark/macro/macro.go
```diff
@@ -134,13 +134,36 @@ func ExtractMacros(
...
+				cfg := map[string]interface{}{}
 
-				return nil
+				err = yaml.Unmarshal([]byte(config), &cfg)
+				if err != nil {
+					err = karma.Format(
+						err,
+						"unable to unmarshal macros config template",
+					)
+					return nil
```
kovetskiy, line 146: should not we return the error here and below? karma.Format only adds a context-like message to an error.

pkg/mark/macro/macro.go
```diff
@@ -134,13 +134,36 @@ func ExtractMacros(
...
+				}
+				body, _ := cfg[template[1:]].(string)
```
kovetskiy, line 148: Will it always (like in 100% cases) go well? It doesn't feel too safe actually. Let's check that there is such a field in cfg first with comma-ok pattern?

pkg/mark/macro/macro.go
Skeeve: I copied the error handing from the error handling already present there. You must know: I'm just a go-beginner and mostly learn by looking at other's code. So I assumed the error handling here https://github.com/kovetskiy/mark/blob/4c812741ac590eb153193fe2c6d228e6845974f7/pkg/mark/macro/macro.go#L138-L142 is correct.

pkg/mark/macro/macro.go
Skeeve: As far as I understand, yes.

template is a string of at least 1 character, which will be '#' due to https://github.com/Skeeve/mark/blob/b7a30d6cf4e3fe0984cea44b4101e512b247fdfb/pkg/mark/macro/macro.go#L137

Also body will either be the inline macro's content or an empty string as, as far as I know, in case of a not-present map element, the result will be the empty string here:

https://github.com/Skeeve/mark/blob/b7a30d6cf4e3fe0984cea44b4101e512b247fdfb/pkg/mark/macro/macro.go#L148

So while it might produce unexpected results, it shouldn't crash.

But yes… An appropriate error message might be more helpful.

pkg/mark/macro/macro.go
kovetskiy: Ah, you are actually right.

PR #199, opened 2022-06-07
fix 142 redone for current master

### LINKED ISSUE #142: Option to disable template parsing on include tag
I want to be able to include for example ansible files. Ansible uses jinja as templating engine which apparently uses the chars to mark variables "{{ variable }}". 

Therefore its currently not possible to include an ansible file with mark as mark tries to parse the file and substitute variables. 

To fix this an option could be added to the include tag to disable parsing.

Skeeve:
Allow change of delimiters when including a template.

This should fix #142

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,8 +68,28 @@ to the template relative to current working dir, e.g.:
 <!-- Include: <path> -->
 ```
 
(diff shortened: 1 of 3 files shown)


kovetskiy:
Thank you!

PR #202, opened 2022-06-08
Update install instructions

datsickkunt:
Just updated the Readme.md
Go emited warning when I was using "go get" and beged for using go install.


kovetskiy:
Thanks!

PR #229, opened 2022-12-02
Allow Title via h1 without any other metadata

mrueg:
Minimal example:
file.md contains
```
# Test
```

`mark --title-from-h1 -p TOKEN --space "MYSPACE" -l URL  -f file.md`

failed before this change with `FATAL specified file doesn't contain metadata and URL is not specified via command line or doesn't contain pageId GET-parameter`

This does not fail for

file2.md
```
<!-- Title: Test -->
```
`mark -p TOKEN --space "MYSPACE" -l URL  -f file2.md`

With this change,  we can add files even if they don't contain metadata but generate the title from their h1 headings

### DIFF
--- a/README.md
+++ b/README.md
@@ -503,10 +503,10 @@ mark -h | --help
 - `-c <path>` or `--config <path>` — Specify a path to the configuration file.
 - `-k` — Lock page editing to current user only to prevent accidental
     manual edits over Confluence Web UI.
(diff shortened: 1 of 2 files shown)


mrueg:
@kovetskiy if you have some spare capacity, could you take a look and provide feedback if that's the right thing to do? Thanks!

PR #230, opened 2022-12-02
Replace deprecated io/ioutils

mrueg:
"Deprecated: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code. See the specific function documentation for details. "


kovetskiy:
Thank you for your contribution!

PR #234, opened 2022-12-13
feat: support ac:jiraissues macro

Nr18:
Add support for the `ac:jiraissues` macro. By adding the template to the stdlib. Not sure if there are other places where you need to add stuff for macros. Or how I could test this locally.


mrueg (approved):

mrueg:
Thanks!

PR #239, opened 2023-01-06
.github: Move CI into the right folder

mrueg:
bde4b702427f488e61d21e6cd7d856dd75fdfefd had the file in the wrong folder.


mrueg:
@kovetskiy thanks for the merge, any suggestion on what to do about 
```
SA1019: strings.Title has been deprecated since Go 1.18 and an alternative has been available since Go 1.0: The rule Title uses for word boundaries does not handle Unicode punctuation properly. Use golang.org/x/text/cases instead.  (staticcheck)
```

I would suggest to add a lint-ignore directive after this one, as there's no 1:1 replacement and it might cause titles to have capitalization changes.

PR #244, opened 2023-01-25
fix: Title not being extracted when not on the first line

### LINKED ISSUE #214: Macro :children: is failing because of missing title
Hi,

I use the following markdown file (test.md) to create a Confluence page:

```
<!-- Space: TestSpace -->
<!-- Macro: :children:
     Template: ac:children
     Style: h2
     Excerpt: none
     Page: TestSpace:Test
     Reverse: false
     All: true -->

# Test

:children:
```

The configuration file that I use:

```
username = "username"
password = "password"
base_url = "url"
h1_title = true
h1_drop = true
```

When I run **mark --title-from-h1 --trace --debug -f test.md** I get the following error:

```
2022-08-23 14:51:44.799 INFO  processing test.md
2022-08-23 14:51:44.799 FATAL page title is not set ('Title' header is not set and '--title-from-h1' option and 'h1_title' config is not set or there is no H1 in the file)
```

As soon as I add the Mark header **** the error message is gone and I receive a 200 response.

I guess this is a bug, because I use the **--title-from-h1** option. Can this be the case, because my configuration file has the **h1_drop** option.

I don't mind to create a pull request and figure out if I can solve this issue, but I don't have any experience with Go and need some help to start.

xiu:
While a page with a macro is processed, having a macro leaves an unclean buffer for further processing steps, including title extraction.  The problem seems to come from `ExtractMeta`: as it runs through the metadata, it hits the first line of the macro but since it doesn't look like a complete Metadata (e.g. matching either `\[\]:\s*#\s*\(([^:]+):\s*(.*)\)` or `<!--\s*([^:]+):\s*(.*)\s*-->` in
https://github.com/kovetskiy/mark/blob/master/pkg/mark/meta.go#L37:L38, it will break the loop in
https://github.com/kovetskiy/mark/blob/master/pkg/mark/meta.go#L61 and return the final document as:
```
  Template: ac:children
  Style: h2
  Excerpt: none
  Page:
  TestSpace:Test
  Reverse: false
  All: true -->

:children:
```

This then goes into `ExtractDocumentLeadingH1` which doesn't find a match for the regex `^#[^#]\s*(.*)\s*\n`, returning an empty title since the title is not on the first line. This first commit, while it doesn't fix the unclean document, fixes the regex to properly find the title.

The second commit makes sure we properly return the macro when hitting one as we process metadata.

Closes #214


README.md
```diff
@@ -120,6 +120,9 @@ be replaced with specified template:
      <yaml-data> -->
 ```
 
+**NOTE**: Make sure to define your macros after your metada (Title/Space), mark
```
mrueg, line 123:
```suggestion
**NOTE**: Make sure to define your macros after your metadata (Title/Space), mark 
```

PR #251, opened 2023-02-06
bugfix: Remove forced default width, forced full-width

jay-stillman:
Updated bugFix due to changes in user editor config - setting default width no longer enables full width publihsing.

### DIFF
--- a/pkg/confluence/api.go
+++ b/pkg/confluence/api.go
@@ -555,16 +555,16 @@ func (api *API) UpdatePage(
 		},
 		"metadata": map[string]interface{}{
 			"labels": labels,
-			// The Confluence API randomly sets page width for some reason:
-			// https://community.developer.atlassian.com/t/confluence-rest-api-create-content-at-random-width/57001
-
-			// This is a workaround to compensate for that bug. It forces the page to be created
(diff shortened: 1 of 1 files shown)


jay-stillman:
fixes #173

mrueg:
Is this change backwards compatible to older versions of confluence?

jay-stillman:
@mrueg full-width is the original and how confluence was prior to the new editor. The "default" mode in latest is based on the new editor fixed-width which is only circa 80% wide. The full-width value is backwards compatible  as far as I have seen, do you have an issue running this? In addition to this have left out content-appearance-draft as that is user preference and configured as a profile setting

jay-stillman:
@mrueg also for clarity in May 22 Atlassian fixed the api for fixed-width, however another bug was introduced. A fix was applied in Dec 22 which resulted in allowing either value to the API. The default is "fixed-width" hence causing the issue with full-width in mark when passing "Layout: plain"

mrueg:
Thanks a lot for the clarification @jay-stillman! I haven't seen any issues here, wanted to make sure that it doesn't break things for other users.

PR #254, opened 2023-02-13
Fixes #253 Can't build with Go versions > 17

### LINKED ISSUE #253: Can't build with Go versions > 17
I get the error
```
error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.
```
when I use the dockerfile to build mark. If you add `-buildvcs=false` to Makefile under build, the error is fixed.

Fethbita:
Fixes the build for go versions >= 18.


mrueg:
I would not disable this option, can we use this recommendation https://github.com/docker-library/golang/issues/452#issuecomment-1423269095 instead?

PR #257, opened 2023-02-27
Use Kroki for mermaid diagrams

tarantoj-intrepid:
Updates the code block template to use [Kroki](https://kroki.io/) for embedding mermaid diagrams, removing the dependency on a confluence extension (paid or otherwise)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*.md eol=lf
+*.html eol=lf
--- a/pkg/mark/stdlib/stdlib.go
+++ b/pkg/mark/stdlib/stdlib.go
@@ -1,6 +1,10 @@
 package stdlib
 
 import (
+	"bytes"
+	"compress/zlib"
+	"encoding/base64"
+	"fmt"
 	"strings"
 	"text/template"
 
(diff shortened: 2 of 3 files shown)


WilsonSunBritten:
I was just checking for mermaid support in this project, great to see this.

mrueg:
Thanks for improving support here!

Wouldn't it be easier to revive https://github.com/kovetskiy/mark/pull/167 instead of relying on a third party service in general?  Also as far as I can see, that would mean the generated images are not pushed to confluence?

tarantoj-intrepid:
@mrueg I did see that pr! It's a cool idea, but for my purposes it was easier and simpler to just use kroki. Definitely adds risk depending on something external, but it was a low effort and low overhead alternative to installing the paid mermaid plugin currently in use. Not an ideal implementation, but personally I think a better interim approach. Also keeps the CI side of things a bit cheaper/quicker

kovetskiy:
@tarantoj-intrepid agree on the idea regarding not using some paid mermaid plugin.

I think we definitely should get rid of the paid one.

We can avoid the issue of the possible data leak mentioned by @mrueg if we will explicitly show that we are going to use a third-party service. i.e. we could use `mermaid-kroki` instead of `mermaid` when specifying "language", so it'll be up to the user which way they want to render it.

I feel like this is third or fourth iteration of mermaid integration, and it could be completely solved if we could render mermaid and upload a file as an attachment. I see there is a package that can render mermaid already: https://pkg.go.dev/github.com/abhinav/goldmark-mermaid

mrueg:
> We can avoid the issue of the possible data leak mentioned by @mrueg if we will explicitly show that we are going to use a third-party service. i.e. we could use `mermaid-kroki` instead of `mermaid` when specifying "language", so it'll be up to the user which way they want to render it.
> 

Another option would be to use something like 

```
```mermaid
my code
```<!-- Provider: kroki -->
```


> I feel like this is third or fourth iteration of mermaid integration, and it could be completely solved if we could render mermaid and upload a file as an attachment. I see there is a package that can render mermaid already: https://pkg.go.dev/github.com/abhinav/goldmark-mermaid

That would be a great solution and probably a good idea since blackfriday doesn't seem to be maintained anymore.

mrueg:
This seems to work reasonably well with mermaid + mermaid.go so I'm closing this one.

PR #258, opened 2023-03-06
Add Content-Appearance header

bernd:
Allows switching between the "full-width" and "fixed" page layouts.

The "fixed" layout renders the page in a narrow column. (Confluence default)

If not configured, it defaults to "full-width".


mrueg:
Thanks!

PR #260, opened 2023-03-15
Replace blackfriday with goldmark

mrueg:
This change replaces the unmaintained blackfriday markdown renderer with the goldmark renderer.
The goal is to have wider support for more markdown extensions and ideally have a better solution for mermaid transformation via extensions as well.


kovetskiy (approved):
LGTM

PR #273, opened 2023-03-31
Implement a custom parser for `<ac:*/>` tags

bernd:
This replaces the workaround to replace colons in `<ac:*/>` tags with a magic string with a custom parser for these tags to parse them as `ast.KindRawHtml`.

The custom parser is a stripped down version of goldmark's rawHTMLParser.

Since we now have a custom parser for `<ac:*/>` tags, the custom link renderer added an additional `</a>` tag at the end of each internal Confluence link.

Add tests for internal links and add an example for internal links with spaces in page titles to the README file.


mrueg:
Thanks a lot!

PR #284, opened 2023-04-04
fix: Remove double interpolation for attachments (BREAKING)

zopanix:
The use case was that when using the "attachment://", it would replace the link twice, once on the "attachment://" and after that on the replace all as well as the string will be present in the replacement. This ensures that the keyword only get replcaed once.

Furthermore, put "attachment://" in as a identifier for a replacement is better then nothing since you effectively create a ban word in the name of the attachement file.

if you have an `attachement://foo`, currently it get transformed into `/wiki/<path>/foo?query` which get transformed into `/wiki/<path>/wiki/<path>/foo?query?query`

this PR changes it to only have a single interpolation.

Note that this is a breaking change as simple text references to files will break. But it seems better to prefix it with `attachement://` anyways to avoid creating ban words in your text.

### DIFF
--- a/pkg/mark/attachment.go
+++ b/pkg/mark/attachment.go
@@ -207,20 +207,6 @@ func CompileAttachmentLinks(markdown []byte, attaches []Attachment) []byte {
 		if bytes.Contains(markdown, []byte("attachment://"+replace)) {
 			from := "attachment://" + replace
 
-			log.Debugf(nil, "replacing legacy link: %q -> %q", from, to)
-
-			markdown = bytes.ReplaceAll(
-				markdown,
-				[]byte(from),
-				[]byte(to),
-			)
-
-			found = true
-		}
-
-		if bytes.Contains(markdown, []byte(replace)) {
-			from := replace
-
 			log.Debugf(nil, "replacing link: %q -> %q", from, to)
 
 			markdown = bytes.ReplaceAll(


kovetskiy:
Hey, could you please elaborate on this point?

> Note that this is a breaking change as simple text references to files will break. But it seems better to prefix it with attachement:// anyways to avoid creating ban words in your text.

zopanix:
Hey, so I'm removing the replacing "<attachment name>" by "<path to attachment name>" which only leaves the possibility to reference it by using "attachment://<attachment name>" to be replaced by <path to attachment name>. This will affect users that are currently counting on automatic replacement of their <attachment filename> by <path to attachement name>

As an example:
```markdown
...
<!-- Attachment: foo.txt -->

As you can read in foo.txt, blablabla
```

Today this becomes (in confluence)
```
As you can read in /wiki/.../foo.txt?query..., blablabla
```

To keep the same behavior users will have to use this:
...


As you can read in attachment://foo.txt, blablabla
```

**Comment by kovetskiy** [comment]
I still don't understand the point of this PR. We are currently considering `attachment://` as legacy and we don't want to use it because it is not supported by markdown viewers such as GitHub.

```
mark -f a.md -u e.kovetskiy@gmail.com -p ... -b https://mark-confluence-dev.atlassian.net/wiki
```

a.md:
```
<!-- Attachment: red.jpg -->
<!-- Attachment: blue.jpg -->
<!-- Title: Test attachments -->
<!-- Space: MARK -->

# red

<img src="attachment://red.jpg" />
<img src="red.jpg" />

# blue

<img src="blue.jpg" />
<img src="attachment://blue.jpg" />
```

Result looks good to me:
https://mark-confluence-dev.atlassian.net/wiki/spaces/MARK/pages/4030465/Test+attachments

cc @mrueg

mrueg:
Thank you for this, and apologies it sat for so long. Both problems you identified are now fixed on master, so I'm closing this — but the diagnosis here was correct and it's worth recording what happened to each.

**The double interpolation.** Exactly as you described: `attachment://foo` became `/wiki/<path>/foo?query`, and then the second replacement matched `foo` inside the URL just written, giving `/wiki/<path>/wiki/<path>/foo?query?query`. The cause was that the legacy `attachment://` branch was an `if` rather than an `else if`, so both branches ran. Fixed in b09a86d.

**The "ban word" problem.** Your other point — that a plain text reference to a filename gets rewritten, effectively making the attachment's name unusable in prose — was also real. `See the file a.png in the repo` became a download URL. That is fixed too: attachment destinations are now rewritten on the parsed document (#888) rather than by substituting text, so only a real link or image destination is touched. Prose, code spans and fenced blocks are left alone.

That last change is why the breaking half of this PR turned out not to be needed. Requiring the `attachment://` prefix would have solved the ban-word problem by making plain references stop working; resolving on the AST solves it while leaving `x` working as before. Both spellings are still accepted.

Regression tests for all of it are in `TestResolver` — including `legacy form is substituted once`, which is your bug specifically.

PR #286, opened 2023-04-18
Simplify config handling

mrueg:
* Switch to urfave/cli/v2
* Add more environment variables


kovetskiy:
While I appreciate the work that has gone into replacing the parsing library, I believe we need to prioritize backwards compatibility in our decision making. It is important to consider the impact that changes will have on our users, and in this case, it seems that the changes would require unnecessary adjustments to their configurations. Can we explore alternative solutions or ways to seamlessly integrate the new library without disrupting our users' workflows?

main.go
```diff
@@ -16,96 +15,168 @@ import (
...
+		Version:     version,
+		Flags:       flags,
+		Before: altsrc.InitInputSourceWithContext(flags,
+			func(context *cli.Context) (altsrc.InputSourceContext, error) {
+				if context.IsSet("config") {
+					filePath := context.String("config")
+					return altsrc.NewTomlSourceFromFile(filePath)
+				} else {
+					// Fall back to default if config is unset
+					return altsrc.NewTomlSourceFromFile(filepath.Join(os.Getenv("HOME"), ".config/mark"))
```
kovetskiy, line 155: I believe here we can try to read data from TOML, and somehow convert under_score keys to arg-like-params. This would lead to no changes in users' workflow.

mrueg: @kovetskiy I've added additional aliases to the flags, which should make it backward compatible.

PR #288, opened 2023-04-24
Add support for mermaid via library

mrueg:
This MR is based on and will replace: https://github.com/kovetskiy/mark/pull/167
Kudos to @dreampuf for the initial implementation

* Adding Mermaid support via mermaid.go, allowing to render it locally.
* Adding new optional arguments to ac:image namely: Title, Heigth attributes
* Switching docker image base to headless-shell


pkg/mark/attachment.go
```diff
@@ -159,14 +155,51 @@ func ResolveAttachments(
...
+
+		attach.Checksum = checksum
+	}
+	return attaches, err
+}
+
+func prepareAttachments(opener vfs.Opener, base string, replacements []string) ([]Attachment, error) {
 	attaches := []Attachment{}
 	for _, name := range replacements {
+		attachmentPath := filepath.Join(base, name)
```
kovetskiy, line 181: this func grew up a bit, worth doing something like this:
```
func prepareAttachments(opener vfs.Opener, base string, replacements []string) ([]Attachment, error) {
    attachments := make([]Attachment, 0, len(replacements))

    for _, name := range replacements {
        attachment, err := createAttachment(opener, base, name)
        if err != nil {
            return nil, err
        }

        attachments = append(attachments, attachment)
    }

    return attachments, nil
}

func createAttachment(opener vfs.Opener, base, name string) (Attachment, error) {
    attachmentPath := filepath.Join(base, name)
    file, err := opener.Open(attachmentPath)
    if err != nil {
        return Attachment{}, karma.Format(err, "unable to open file: %q", attachmentPath)
    }
    defer file.Close()

    fileBytes, err := io.ReadAll(file)
    if err != nil {
        return Attachment{}, karma.Format(err, "unable to read file: %q", attachmentPath)
    }

    return Attachment{
        Name:      name,
        Filename:  strings.ReplaceAll(name, "/", "_"),
        FileBytes: fileBytes,
        Replace:   name,
    }, nil
}
```

mrueg: I added a resolveAttachement function.

Issue #305, opened 2023-05-05
External ac:image URLs are not escaped properly 


recrtl:
**What happened?**
Confluence is throwing errors when images with multiple query parameters, because `&` characters are not escaped to `&amp;` (see https://jira.atlassian.com/browse/CONFSERVER-35200).
This is happening since the update that uses ac:image for images https://github.com/kovetskiy/mark/commit/fd97ee70f96d0a63d08297c4d0d1fb8424ebc589
When uploading the markdown file, Confluence complains with a 400 error.

**What did you expect to happen?**
External images URLs are escaped and Confluence accepts them.
To do that, this .md line
`![My External Image](http://confluence.atlassian.com/images/logo/confluence_48_trans.png?key1=value1&key2=value2)`
Should be transformed to 
`<p><ac:image ac:alt="My External Image"><ri:url ri:value="http://confluence.atlassian.com/images/logo/confluence_48_trans.png?key1=value1&amp;key2=value2"/></ac:image></p>`

**How can we reproduce the behavior you experienced?**
1. Include a .md file with an external image URL with multiple query parameters, e.g. `![My External Image](http://confluence.atlassian.com/images/logo/confluence_48_trans.png?key1=value1&key2=value2)`
2. Run mark and upload to Confluence
3. Confluence returns a 404

** Information (please complete the following information):**
 - Mark Version (`mark --version`): 3.9.0
 - Mark Parameters: n/a
 - Confluence Version: v7.18.1
 - Environment specific Information: running locally or in github actions

** Logs or other output**
```
FATAL Confluence API returned unexpected status: 400 Bad Request, output: "{\"statusCode\":400,\"data\":{\"authorized\":false,\"valid\":true,\"allowedInReadOnlyMode\":true,\"errors\":[],\"successful\":false},\"message\":\"Error parsing xhtml: Unexpected character '=' (code 61); expected a semi-colon after the reference for entity 'key2'\\n at [row,col {unknown-source}]: [11,212]\",\"reason\":\"Bad Request\"}"
```

**Additional context**
N/A

recrtl:
Proposing fix in PR https://github.com/kovetskiy/mark/pull/306

PR #306, opened 2023-05-05
fix 305: external images URLs are not properly escaped

recrtl:
fixes https://github.com/kovetskiy/mark/issues/305


mrueg:
Good catch, Thanks a lot!

PR #318, opened 2023-05-19
Move drop-h1 to the renderer

mrueg:
This is more reliable than a regular expression.

@kovetskiy do you know if we want to drop *ALL* h1 or just the first one with this?


mrueg:
Testing it out, it looks like it'd only drop the first h1.

kovetskiy:
> @kovetskiy do you know if we want to drop ALL h1 or just the first one with this?

Sorry for late reply, we need to drop only the first one. This is used by people who render title of the page with h1 (`# header`) on GitHub (or any other markdown reader), but when they submit the page to Confluence, they would have see two titles -- one from confluence and one from h1. Users came with with a flag to drop only the first h1.

PR #335, opened 2023-07-03
doc: clarify links usage of page title with spaces

gmarraff:
The word space can be misleading in this context since it can be confused as 'Confluence Space'.

It happened to me.


mrueg:
Thanks for the PR!

Issue #324, opened 2023-05-24
Table within macro is not rendered


mrueg:
**What happened?**
Trying to render this to confluence
```
<ac:structured-macro ac:name="info">
<ac:parameter ac:name="icon">true</ac:parameter>
<ac:parameter ac:name="title">Attention</ac:parameter>
<ac:rich-text-body>| Header 1 | Header 2 |
|---|---|
| Cell A | Cell B |
</ac:rich-text-body>
</ac:structured-macro>
```

results in 

```
                                <p><ac:structured-macro ac:name="info">
                                <ac:parameter ac:name="icon">true</ac:parameter>
                                <ac:parameter ac:name="title">Attention</ac:parameter>
                                <ac:rich-text-body></p>
                                <table>
                                <thead>
                                <tr>
                                <th>Header 1</th>
                                <th>Header 2</th>
                                </tr>
                                </thead>
                                <tbody>
                                <tr>
                                <td>Column A</td>
                                <td>Column B</td>
                                </tr>
                                </tbody>
                                </table>
                                <p></ac:rich-text-body>
                                </ac:structured-macro></p>

```

**What did you expect to happen?**
Table gets rendered

```
<ac:structured-macro ac:name="info">
<ac:parameter ac:name="icon">true</ac:parameter>
<ac:parameter ac:name="title">Attention</ac:parameter>
<ac:rich-text-body><table>
<thead>
<tr>
<th>Header 1</th>
<th>Header 2</th>
</tr>
</thead>
<tbody>
<tr>
<td>Cell A</td>
<td>Cell B</td>
</tr>
</tbody>
</table>
</ac:rich-text-body>
</ac:structured-macro>
```

**How can we reproduce the behavior you experienced?**
1. Use markdown above
2. Render it with mark
3. 
In case this is related to specific markdown, please provide a minimal markdown example here.

** Information (please complete the following information):**
 - Mark Version (`mark --version`): [e.g. v9.1.4]  9.5.2
 - Mark Parameters: [e.g. --drop-h1 --title-from-h1]
 - Confluence Version: [e.g. v7.13]
 - Environment specific Information: [e.g. running in Github Actions, or on Mac OS X, etc.]

** Logs or other output**
Please provide logs, other kind of output or observed metrics here.

**Additional context**

mrueg:
[...]

PR #351, opened 2023-09-06
Adding linebreak stripping

### LINKED ISSUE #92: Handle semantic line breaks when converting markdown file to an HTML file

nyarly:
Adds `--strip-newlines` which collapses single newlines into a space - effectively what would happen during normal HTML rendering.  This is to address Confluence's behavior, which is to insert a `<br>`.

Closes #92

### DIFF
--- a/main.go
+++ b/main.go
@@ -64,6 +64,13 @@ var flags = []cli.Flag{
 		EnvVars: []string{"MARK_H1_DROP"},
 	}),
 	altsrc.NewBoolFlag(&cli.BoolFlag{
+		Name:    "strip-linebreaks",
+		Value:   false,
+		Aliases: []string{"L"},
+		Usage:   "remove linebreaks inside of tags, to accomodate non-standard Confluence behavior",
+		EnvVars: []string{"MARK_STRIP_LINEBREAK"},
+	}),
+	altsrc.NewBoolFlag(&cli.BoolFlag{
 		Name:    "title-from-h1",
 		Value:   false,
 		Aliases: []string{"h1_title"},
@@ -383,7 +390,7 @@ func processFile(
 			)
 		}
 
-		html, _ := mark.CompileMarkdown(markdown, stdlib, file, cCtx.String("mermaid-provider"), cCtx.Float64("mermaid-scale"), cCtx.Bool("drop-h1"))
+		html, _ := mark.CompileMarkdown(markdown, stdlib, file, cCtx.String("mermaid-provider"), cCtx.Float64("mermaid-scale"), cCtx.Bool("drop-h1"), cCtx.Bool("strip-linebreaks"))
 		fmt.Println(html)
 		os.Exit(0)
 	}
@@ -459,7 +466,7 @@ func processFile(
 		)
 	}
 
-	html, inlineAttachments := mark.CompileMarkdown(markdown, stdlib, file, cCtx.String("mermaid-provider"), cCtx.Float64("mermaid-scale"), cCtx.Bool("drop-h1"))
+	html, inlineAttachments := mark.CompileMarkdown(markdown, stdlib, file, cCtx.String("mermaid-provider"), cCtx.Float64("mermaid-scale"), cCtx.Bool("drop-h1"), cCtx.Bool("strip-linebreaks"))
 
 	// Resolve attachements detected from markdown
 	_, err = attachment.ResolveAttachments(
--- a/pkg/mark/markdown_test.go
+++ b/pkg/mark/markdown_test.go
@@ -10,6 +10,24 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func loadData(t *testing.T, filename, variant string) ([]byte, string, []byte) {
+	t.Helper()
+	basename := filepath.Base(filename)
+	testname := strings.TrimSuffix(basename, ".md")
+	htmlname := filepath.Join(filepath.Dir(filename), testname+variant+".html")
+
+	markdown, err := os.ReadFile(filename)
+	if err != nil {
+		panic(err)
+	}
+	html, err := os.ReadFile(htmlname)
+	if err != nil {
+		panic(err)
+	}
+
+	return markdown, htmlname, html
+}
+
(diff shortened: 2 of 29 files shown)


pkg/mark/markdown_test.go
```diff
@@ -19,24 +37,50 @@ func TestCompileMarkdown(t *testing.T) {
...
 			panic(err)
 		}
-		html, err := os.ReadFile(htmlname)
+		markdown, htmlname, html := loadData(t, filename, "")
+		actual, _ := CompileMarkdown(markdown, lib, filename, "", 1.0, false, false)
+		test.EqualValues(string(html), actual, filename+" vs "+htmlname)
+	}
+}
+
+func TestCompileMarkdownDropH1(t *testing.T) {
```
nyarly, line 50: Also added tests for DropH1 - I can't swear to their accuracy, but they look right

pkg/mark/renderer/text.go
```diff
@@ -0,0 +1,71 @@
...
+						thisLastRune := util.ToRune(value, len(value)-1)
+						siblingFirstRune, _ := utf8.DecodeRune(siblingText)
+						if !(util.IsEastAsianWideRune(thisLastRune) &&
+							util.IsEastAsianWideRune(siblingFirstRune)) {
+							_ = w.WriteByte(byte(r.softBreak))
+						}
+					}
+				}
+			} else {
+				_ = w.WriteByte(byte(r.softBreak))
```
nyarly, line 66: This is effectively taken directly from goldmark (like the other renderers), with this exception.

pkg/mark/testdata/newlines-stripnewlines.html
```diff
@@ -0,0 +1,8 @@
+<p>one-1 one-2</p>
+<p>two-1</p>
+<p>two-2</p>
+<p>three-1</p>
+<p>three-2</p>
+<p>space-1 space-2</p>
```
nyarly, line 6: This is what we want

pkg/mark/testdata/quotes-stripnewlines.html
```diff
@@ -0,0 +1,33 @@
...
+<h2 id="second-heading">Second Heading</h2>
+<ac:structured-macro ac:name="warn"><ac:parameter ac:name="icon">true</ac:parameter><ac:rich-text-body>
+<p><strong>Warn</strong></p>
+<ul>
+<li>Warn bullet 1</li>
+<li>Warn bullet 2</li>
+</ul>
+</ac:rich-text-body></ac:structured-macro>
+<ul>
+<li>Regular list that runs long</li>
```
nyarly, line 23: This is the thing that cannot be (easily) got by e.g. a sed script.

pkg/mark/markdown_test.go
mrueg: Thanks for adding these!

pkg/mark/renderer/text.go
```diff
@@ -0,0 +1,71 @@
+package renderer
+
+import (
+	"unicode/utf8"
+
+	//"github.com/kovetskiy/mark/pkg/mark/stdlib"
```
mrueg, line 6: Can you remove the import here if you don't need it?

mrueg:
Thanks for the contribution! I've added a couple of comments to it.

main.go
```diff
@@ -63,6 +63,13 @@ var flags = []cli.Flag{
 		Usage:   "don't include the first H1 heading in Confluence output.",
 		EnvVars: []string{"MARK_H1_DROP"},
 	}),
+	altsrc.NewBoolFlag(&cli.BoolFlag{
+		Name:    "strip-linebreaks",
+		Value:   false,
```
mrueg, line 68: Should we enable it by default?

pkg/mark/renderer/text.go
mrueg: Can you add a reference where you took it from as a comment?

pkg/mark/renderer/text.go
```diff
@@ -0,0 +1,71 @@
...
+		softBreak: sb,
+	}
+}
+
+// RegisterFuncs implements NodeRenderer.RegisterFuncs .
+func (r *ConfluenceTextRenderer) RegisterFuncs(reg renderer.NodeRendererFuncRegisterer) {
+	reg.Register(ast.KindText, r.renderText)
+}
+
+func (r *ConfluenceTextRenderer) renderText(w util.BufWriter, source []byte, node ast.Node, entering bool) (ast.WalkStatus, error) {
```
mrueg, line 35: Maybe add a comment on top why this is needed and what it does instead of the default text renderer from goldmark

main.go
nyarly: It would be a change to the output, which is why I set it disabled by default. If it were me, I'd enable by default on the next major release, and remove the option on the release after that.

mrueg: Okay, makes sense, I'll switch it on once we release mark-10.x

mrueg:
Thanks for the feature @nyarly ! While I have you here and you probably also looked a bit deeper into goldmark, any thoughts on how to solve https://github.com/kovetskiy/mark/issues/324 ?

PR #354, opened 2023-09-07
Feature/absolute path

Skeeve:
I think it could be handy to have globel includes.

But as the path has to be relative to the current working directory, I created this patch:

- A relative path is relative to the current working directory. Same as before
- An absolute path (staring with /) is taken as-is.
- A path starting with ~/ starts at the user's home directory.
- A path starting with a double slash // starts at the directory mark.d in the user's config directory.


mrueg:
I think this feature needs a bit more discussion.

The `//` pretty custom and not intuitive. Allowing file access outside the base directory could also cause undesired access patterns (e.g. being able to print the mark config that includes the credentials to confluence) into confluence. In particular in CI pipelines this could be a risk.

In the next couple of months I want to move all the custom logic (e.g. Macros, Includes, etc.) into the goldmark extension as well to avoid having two different places for file manipulation.

Skeeve:
You can already access the config file to print the credentials when you know its position in relation to the current directory. So I do not see any security gain here.

Regarding `//` I simply used it because
a. It's normally nothing a path starts with.
b. I had no other idea, but I would be happy to adopt any other reasonable proposal.

Skeeve:
What about this idea:
1. Drop absolut paths
2. Drop the ~
3. Only allow relative paths
4. So everything as is right now, but: If an included file cannot be found, search it below $CONFIG/mark.d

PR #360, opened 2023-09-26
Bugfix/tablealignment

Skeeve:
Fixes the table alignment issues by adding some CSS to every page.


mrueg:
From a first glance, the API doesn't seem to be the right place for those changes.

Skeeve:
> From a first glance, the API doesn't seem to be the right place for those changes.

Can you please elaborate?

Maybe there is a better place for that CSS?

mrueg:
The appropriate place, if this alignment is not properly supported by https://github.com/yuin/goldmark's table extension would be
* The table extension from goldmark itself
*  A customization of the extension in the renderer  https://github.com/kovetskiy/mark/tree/master/pkg/mark/renderer

Skeeve:
goldmark you cannot change, as it is following XHTML standards, as stated in their repository. So the align-attribute is correct there. But it violates HTML5.

For the renderer, t.b.h. I have no clue yet how to do it.

mrueg:
Replaced by #376

PR #375, opened 2023-10-17
feat: Add fallback directory to load includes from

mrueg:
This is a more flexible solution than https://github.com/kovetskiy/mark/pull/355 and allows users to set custom include paths as a fallback.

@Skeeve please take a look and let me know if this works for you as well. You should be able to use it via `--include-path "~/config/mark.d"` or similar.


Skeeve:
Without testing: Fine for me.
I assume you can set the path in the `mark`-file.

PR #376, opened 2023-10-17
change: Drop XHTML

mrueg:
Use HTML5 rendering instead 

This should solve #360 by creating style attributes. https://github.com/yuin/goldmark/blob/master/extension/table.go#L502

@Skeeve


Skeeve:
Will test.

mrueg:
Maybe you tested with a different binary? I've added some tests for aligned tables and they use the style attribute.

Skeeve:
Maybe I did something wrong, but I thought I checked out [mrueg:drop-xhtml](https://github.com/mrueg/mark/tree/drop-xhtml)

Anyhow: Upon a new test, it worked as described.

PR #392, opened 2023-11-24
feat: Support for double star (**) globbing

### LINKED ISSUE #105: Support double star (**) glob syntax

oradwell:
Very annoying that the go [filepath](https://pkg.go.dev/path/filepath#Match) package doesn't support double star globbing. Using the [bmatcuk/doublestar](https://github.com/bmatcuk/doublestar) package as recommended in https://github.com/golang/go/issues/11862

Fixes #105

### DIFF
--- a/main.go
+++ b/main.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/bmatcuk/doublestar/v4"
 	"github.com/kovetskiy/lorg"
 	"github.com/kovetskiy/mark/pkg/confluence"
 	"github.com/kovetskiy/mark/pkg/mark"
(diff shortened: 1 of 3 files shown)


jimmcslim:
Would be great if this was merged, will simplify some GitHub workflows we have written to publish documentation to Confluence!

main.go
```diff
@@ -239,7 +240,7 @@ func RunMark(cCtx *cli.Context) error {
 
 	api := confluence.NewAPI(creds.BaseURL, creds.Username, creds.Password)
 
-	files, err := filepath.Glob(cCtx.String("files"))
+	files, err := doublestar.Glob(os.DirFS("."), cCtx.String("files"))
```
mrueg, line 243: Why not use https://github.com/bmatcuk/doublestar#filepathglob ?

oradwell:
```suggestion
	files, err := doublestar.FilepathGlob(cCtx.String("files"))
```

oradwell: thanks for the suggestion. I've done that change. FilepathGlob is definitely better

oradwell:
@mrueg thanks for merging. when should we expect a release containing this feature?

PR #401, opened 2023-12-18
Fix layout section attribute tags

stephenpaulger:
I was finding that without `ac:` prefix on the `type` attribute that the attribute was being ignored.

I wondered if the attribute had been expected without the prefix in other versions of confluence but is appears this is correct in all the versions currently documented.

- [8.7](https://confluence.atlassian.com/doc/confluence-storage-format-790796544.html#ConfluenceStorageFormat-Pagelayouts)
- [8.0](https://confluence.atlassian.com/conf80/confluence-storage-format-1188407674.html#ConfluenceStorageFormat-Pagelayouts)
- [7.16](https://confluence.atlassian.com/conf716/confluence-storage-format-1108682045.html#ConfluenceStorageFormat-Pagelayouts)

Elsewhere in the mark codebase the `ac:` prefix used too. I've only tested this on a single confluence version but it seems like it should be correct for other versions.

### DIFF
--- a/pkg/mark/renderer/htmlblock.go
+++ b/pkg/mark/renderer/htmlblock.go
@@ -45,22 +45,22 @@ func (r *ConfluenceHTMLBlockRenderer) renderHTMLBlock(w util.BufWriter, source [
 			_, _ = w.WriteString("</ac:layout>\n")
 			return ast.WalkContinue, nil
 		case "<!-- ac:layout-section type:single -->":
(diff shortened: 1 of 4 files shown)


mrueg:
Thanks! I wonder how it worked for me before. :confused:

stephenpaulger:
@mrueg I'm not sure but at a guess some versions of confluence might use a different XML parser that will make different assumptions about attribute prefixes. I only tested it on a quite old version.

PR #405, opened 2024-01-04
container file: only install git-core in container

gthieleb:
This PR helps to only install the minimal git client without X11 dependencies which reduces the size of the Container Image and lowers the number of security CVE reported.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ RUN make get \
 
 FROM chromedp/headless-shell:latest
 RUN apt-get update \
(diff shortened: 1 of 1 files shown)


mrueg:
As far as I understand https://packages.debian.org/bullseye/git-core is a virtual package for https://packages.debian.org/bullseye/git how would this make a difference?

gthieleb:
You are correct. I was the assumption that this package installs the x11 dependencies of git but it seems I was mixing it with git on centos. Sorry for the noise.

PR #420, opened 2024-02-06
Build multi-arch image for amd64 and arm64

stephenpaulger:
I have been building my own multi-arch image locally and that works for me but with this modification the docker image built by the CI and pushed to [docker hub](https://hub.docker.com/r/kovetskiy/mark) will start supporting amd64 and arm64.

If you want to try this locally it is equivalent to building with

```
docker build --platform=linux/amd64,linux/arm64 -t kovetskiy/mark:$tag .
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,6 +107,7 @@ jobs:
       if: ${{ github.ref_type == 'tag' }}
       with:
         push: true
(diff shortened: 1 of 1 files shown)


mrueg:
thanks!

stephenpaulger:
Screenshot 2024-04-12 at 09 38 55

Thanks @mrueg

PR #421, opened 2024-02-08
Pin Github actions dependencies

stephenpaulger:
I've made this change to try to improve mark's OpenSSF ScoreCard. ScoreCard has a check for pinned dependencies, including those used in the CI workflow. Currently mark has 0/10 for pinned dependencies and a contributor to that is not having used pinned actions. A good ScoreCard score can be helpful to users in enterprise environments where there is a need to show that thought has been given to the security of tools being used.

 - [Docs from ScoreCard about pinned dependencies](https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies)
 -  [mark's scorecard scores on deps.dev](https://deps.dev/go/github.com%2Fkovetskiy%2Fmark)

mark is already using dependabot and it looks to be configured to update github actions, it understands this format and will continue to send PRs to update dependencies in the CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
       id: go
 
     - name: golangci-lint
-      uses: golangci/golangci-lint-action@v3
+      uses: golangci/golangci-lint-action@3cfe3a4abbb849e10058ce4af15d205b6da42804 # v3.7.1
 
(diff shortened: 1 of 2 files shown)


stephenpaulger:
This is out of date now, if you'd like to consider doing this you can let me know and I can redo it.

stephenpaulger:
Closing this for now.

PR #439, opened 2024-04-10
Add option to pass a version message

stephenpaulger:
Adds a new option to mark that allows the edit message to be set.

```
   --version-message value         add a message to the page version, to explain the edit (default: "") [$MARK_VERSION_MESSAGE]
```

This value can be passed to mark with the new argument `--version-message` or by setting the environment variable `MARK_VERSION_MESSAGE`.

My use case for this feature is allowing the version information for a collection of confluence pages to be added to the page history in confluence.


mrueg:
Thanks, that's a neat feature!

Issue #448, opened 2024-05-30
template ac:jiraissues returns error 500


ChrisRu82:
## What happened?
If I use the template `ac:jiraissues` I receive from the confluence an Error 500 "Internal Server Error". I tested this with only the required parameters and with all optional parammeters

## What did you expect to happen?

That the Jira issue List would have been created on the confluence page.

## How can we reproduce the behavior you experienced?

1. Add a macro with these values in the markdown

```Markdown
<!--` Macro: :testlist:
     Template: ac:jiraissues
     URL: https://jira.example.com/sr/jira.issueviews:searchrequest-xml/100000/SearchRequest-100000.xml?tempMax=1000
      -->

:testlist:

```

2. run mark

## Further Information (please complete the following information)

* Mark Version (`mark --version`): [ v9.12]
* Mark Parameters: [`--drop-h1 --title-from-h1`]
* Confluence Hosting: [Datacenter]
* Confluence Version: [9.5.4]
* Environment specific Information: [tested in Gitlab CI and Mac OS terminal]

## Logs or other output

Please provide logs, other kind of output here.

## Additional context

The problem can be solved when we change [Line 170](https://github.com/kovetskiy/mark/blame/94a00c5e22ae0bf616b0042f54475a33cea0cbcc/pkg/mark/stdlib/stdlib.go#L170) and 
 [Line 177](https://github.com/kovetskiy/mark/blame/94a00c5e22ae0bf616b0042f54475a33cea0cbcc/pkg/mark/stdlib/stdlib.go#L177) in the `stdlib.go`
 
The missing bit is that the urls need to be in the `ri:url` format similiar to [Line 254](https://github.com/kovetskiy/mark/blame/94a00c5e22ae0bf616b0042f54475a33cea0cbcc/pkg/mark/stdlib/stdlib.go#L254).

mrueg:
Thanks for the PR, I merged it!

PR #449, opened 2024-05-30
fix: ac:jiraissues to use correct format for the urls

ChrisRu82:
This MR fix the issue with the wrong format of the URL for `ac:jiraissues` which is raised in  #448 

Not sure if I need to adjust any tests


mrueg:
Thanks, good catch!

Issue #226, opened 2022-11-09
Label appear and disappear (PingPong)


ahmedloai:
When using the meta-data  I noticed the following behaviour. 

- if the page does not has the label it is added correctly
- When i republish the same page again which already has a label A, it disappeared. 

Please, have check i could be an easy fix Bug

guipaivanz:
This issue has been open for a year and a half. Is there any update?

mrueg:
This might be resolved by #454 please report back if not.

PR #454, opened 2024-06-11
Change label management to rely on independent rest calls

camielwnzl:
Calc delta for labels from metadata and page information, then add/remove labels via independent confluence api calls

Seems to address "Label appear and disappear (PingPong) #226"

### DIFF
--- a/pkg/confluence/api.go
+++ b/pkg/confluence/api.go
@@ -73,6 +73,15 @@ type AttachmentInfo struct {
 	} `json:"_links"`
 }
 
+type Label struct {
+	ID     string `json:"id"`
+	Prefix string `json:"prefix"`
+	Name   string `json:"name"`
+}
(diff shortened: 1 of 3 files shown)


pkg/confluence/api.go
```diff
@@ -514,17 +523,6 @@ func (api *API) UpdatePage(page *PageInfo, newContent string, minorEdit bool, ve
 		}
 	}
 
-	labels := []map[string]interface{}{}
-	for _, label := range newLabels {
-		if label != "" {
-			item := map[string]interface{}{
-				"prexix": "global",
```
mrueg, line 521: looks like there was a typo here anyway. 

Is the change backwards compatible with this one in mind?

mrueg:
Let's see if this improves the situation with labels :) Thanks!

Issue #456, opened 2024-06-14
Accessing page links through a different URL than `base-url` fails


monti-python:
## What happened?

In our company we access the same Confluence instance through different domains depending on the authentication scheme. However, `mark` hardcodes the entire URL (including the `base-url`) whenever a Confluence page from the same instance is referenced via relative paths, creating a hard reference to the `base-url` that was specified at the time `mark` was invoked.

This can be fixed by removing the hardcoded reference to `base-url` as Confluence supports [relative links](https://confluence.atlassian.com/doc/links-776656293.html) to refer to pages under the same Confluence instance.

This would also make the pages and links independent from the current `base-url` in general, allowing for more flexibility.

## What did you expect to happen?

Links to Confluence pages within the same page should always work regardless of the Confluence's hostname/domain.

## How can we reproduce the behavior you experienced?

1. Create a page referencing a relative file containing another page, e.g.:

    ```html
    <!-- Title: PageWithRef -->

    See this [page](../path/to/other/page.md)
    ```
2. Create a DNS alias for your Confluence instance
3. Deploy the page above using the new alias as `base-url`
4. Observe how the references above points to the new `base-url` (DNS alias), even in the regular URL

## Further Information (please complete the following information)

* Mark Version (`mark --version`): 9.12.0

monti-python:
Solved in #457

monti-python:
Reopening due to #468 #469

monti-python:
@mrueg @st-karol As a simple mechanism to form relative links we could simply extract the path from the `base_url`, e.g.:

```go
link = url.Parse(api.BaseURL).Path + page.Links.Full
```

That should be robust enough to handle both Confluence Cloud (with the `/wiki` subpath) and Confluence Server (without a subpath), assuming the `base_url` includes the `/wiki` subpath...

What do you think?

st-karol:
@monti-python 
```go
	if page != nil {
		var path, err = url.Parse(api.BaseURL)
		if nil == err {
			link = path.Path + page.Links.Full
		}
	}
```

Works perfectly fine in our case (cloud). Correct relative link is generated 💥 Go for it if that works with your setup as well 🙇

mrueg:
Sure, feel free to propose the suggestion as a PR!

monti-python:
@st-karol it'd be great if you could have a look at #477 as well 🙏

mrueg:
Closing as #477 has been merged. Thanks for your contribution!

PR #457, opened 2024-06-14
Relative references to pages within the same Confluence instance

monti-python:
Solves #456


mrueg:
Thanks, makes sense to me!

PR #467, opened 2024-07-18
feat: add placeholder in plain layout

lukasriegel:
Support the ac:placeholder in plain layout


mrueg:
That will mean placeholders will not be rendered in markdown but it will show up in confluence. Is that intended behavior or should we rather use a macro instead?

PR #469, opened 2024-07-25
Revert "Make references to pages within the same Confluence relative"

mrueg:
This reverts commit ecd563e0957c23573bc8b37ba398396b8144c0d0.

@monti-python it looks like #457 broke some cloud instances of confluence.

See also #468

### DIFF
--- a/pkg/mark/link.go
+++ b/pkg/mark/link.go
@@ -203,9 +203,9 @@ func getConfluenceLink(
 	}
 
 	if page != nil {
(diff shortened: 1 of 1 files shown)


mrueg:
@monti-python reverting this change for now.
happy to accept a solution that works for confluence cloud and server if this is something you need.

monti-python:
Thanks for the reference @mrueg, I followed up on this in [#456](https://github.com/kovetskiy/mark/issues/456#issuecomment-2251544763)

Issue #466, opened 2024-07-15
ac:details usage issues


OnceUponALoop:
## What happened?

I'm trying to use the `ac:details` template to populate page properties but i'm unable to due to errors in parsing.

`com.atlassian.confluence.api.service.exceptions.BadRequestException: Content body cannot be converted to new editor format`

## What did you expect to happen?

I expected a page properties element with a table of key value pairs

![image](https://github.com/user-attachments/assets/6b33ea8e-e5e7-4e50-b40f-5e82bfb34aa4)

## How can we reproduce the behavior you experienced?

Using the following example

```
<!-- Space: ~01234567 -->
<!-- Parent: Test -->
<!-- Title: Testing Mark -->

<!-- Macro: :details_start:([^:][\s\S]*?):details_end:
     Template: ac:details
     Body: ${1}
-->

:details_start:
| Key1 | Key2 |
| --- | --- |
| Value1 | Value 2 |

:details_end:
```

## Further Information (please complete the following information)

* Mark Version (`mark --version`): v9.12.0
* Mark Parameters: N/A
* Confluence Hosting:  Cloud
* Confluence Version: Cloud
* Environment specific Information: Running in Windows

## Logs or other output

Markdown parsing section of the log

```
2024-07-15 18:07:18.471 TRACE rendering markdown:



                              <ac:structured-macro ac:name="details" ac:schema-version="1""><ac:rich-text-body>
                              | Key1 | Key2 |
                              | --- | --- |
                              | Value1 | Value 2 |

                              </ac:rich-text-body></ac:structured-macro>


2024-07-15 18:07:18.471 TRACE rendered markdown to html:
                              <p>&lt;ac:structured-macro ac:name=&quot;details&quot; ac:schema-version=&quot;1&quot;&quot;&gt;<ac:rich-text-body></p>
                              <table>
                              <thead>
                              <tr>
                              <th>Key1</th>
                              <th>Key2</th>
                              </tr>
                              </thead>
                              <tbody>
                              <tr>
                              <td>Value1</td>
                              <td>Value 2</td>
                              </tr>
                              </tbody>
                              </table>
                              <p></ac:rich-text-body></ac:structured-macro></p>
```

## Additional context

I'm unclear if my `ac:details` usage is correct.

[...]

PR #471, opened 2024-07-25
Do not render paragraph tags

mrueg:
If we stop rendering paragraphs, then macros that surround text will actually work.

Likely a breaking change and will bump the package to v10.


mrueg:
@OnceUponALoop this likely help with #466 as well.

Issue #461, opened 2024-06-27
Links with square brackets are not correctly parsed


ProvoK:
## What happened?

a link like
```
[example [text]](https://www.google.it)
```
won't be parsed correctly by the regexp

## What did you expect to happen?

Expected to correctly parse the link, with the text including the square brackets

## How can we reproduce the behavior you experienced?

1. Add a new test into `pkg/mark/link_test.go` TestParseLink with example below
2. Assert new link parsed correctly
3. test panics (index out of bounds)

Example markdown
```
[example [text]](https://www.google.it)
```

## Further Information (please complete the following information)

* Mark Version (`mark --version`): latest, main

PR #472, opened 2024-07-25
Fix Regex for links that contain square brackets in the text

mrueg:
Should fix https://github.com/kovetskiy/mark/issues/461
@provok


ProvoK:
Awesome thanks!

PR #477, opened 2024-07-27
Use relative links for objects within the same Confluence instance

monti-python:
Solves #456

### DIFF
--- a/pkg/mark/link.go
+++ b/pkg/mark/link.go
@@ -203,10 +203,15 @@ func getConfluenceLink(
 	}
 
 	if page != nil {
-		// Needs baseURL, as REST api response URL doesn't contain subpath ir
-		// confluence is server from that
 		link = api.BaseURL + page.Links.Full
 	}
 
-	return link, nil
+	linkUrl, err := url.Parse(link)
+	if err != nil {
+		return "", karma.Format(err, "parse URL: %s", link)
+	}
+	// Confluence supports relative links to reference other pages:
+	// https://confluence.atlassian.com/doc/links-776656293.html
+	linkPath := linkUrl.Path
+	return linkPath, nil
 }


pkg/mark/link.go
```diff
@@ -203,10 +203,15 @@ func getConfluenceLink(
...
 	}
 
-	return link, nil
+	linkUrl, err := url.Parse(link)
+	if err != nil {
+		return "", karma.Format(err, "parse URL: %s", link)
+	}
+	// Confluence supports relative links to reference other pages:
+	// https://confluence.atlassian.com/doc/links-776656293.html
+	linkPath := linkUrl.Path
```
mrueg, line 215: Won't this always provide the relative path instead of the absolute path with baseurl now? 

I think it would be good to have tests for this one, I'd like to avoid breaking things again. Would you be willing to write some?

monti-python: Hi @mrueg,

Thanks a lot for reviewing!

Yup, relative paths should always work. The problem last time was that, in the case of Confluence Cloud, `baseURL` already contains a subpath that should be prepended to `pageLink`.

In this case we take a more robust by forming the absolute URL first, then returning the relative path to make the links independent from the hostname.

#### Scenario 1 - `baseURL` doesn't contain a subpath (Confluence Server)
- `baseURL`: `https://[my_server_hostname]>`
- `pageLink`: `/spaces/~spaceId/my_page`
- `absoluteURL`: `https://[my_server_hostname]>/spaces/~spaceId/my_page`
- `relativePath`: `/spaces/~spaceId/my_page`

#### Scenario 2 - `baseURL` contains a subpath (Confluence Cloud)
- `baseURL`: `https://[company-name].atlassian.net/wiki`
- `pageLink`: `/spaces/~spaceId/my_page`
- `absoluteURL`: `https://[company-name].atlassian.net/wiki/spaces/~spaceId/my_page`
- `relativePath`: `/wiki/spaces/~spaceId/my_page`

I thought about adding a separate function to isolate and test this functionality:

```go
func getRelativePath(absoluteLink string) (string, error) {
	linkUrl, err := url.Parse(absoluteLink)
	if err != nil {
		return "", karma.Format(err, "parse URL: %s", link)
	}
	return linkUrl.Path, nil
}
```

However, testing it would be equivalent to testing the `url` library (`linkUrl.Path`).

In any case, I'm happy to add those tests if you think they'd provide some value.

mrueg: Thanks for the further explanation. I think this looks good now. I got confused because I initially thought we'll only replace links in cloud. :)

PR #489, opened 2024-08-20
added update-if-older-than option

danapsimer:
This option will skip files where the latest version is have been updated since the date provided on the command line.

For example this could be used to update only if the last commit for a file is newer than the latest version in confluence:

`mark --space "****" -f $file -u "${MARK_USER}" -p "${MARK_PWD}" -b "${MARK_URL}" --ci --update-if-older-than "$(git --no-pager log -1 --pretty="format:%cI" $file)"`

### DIFF
--- a/main.go
+++ b/main.go
@@ -25,7 +25,7 @@ import (
 )
 
 const (
-	version     = "10.0.0"
+	version     = "10.0.1"
 	usage       = "A tool for updating Atlassian Confluence pages from markdown."
 	description = `Mark is a tool to update Atlassian Confluence pages from markdown. Documentation is available here: https://github.com/kovetskiy/mark`
 )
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -25,7 +25,7 @@ import (
 )
 
 const (
-	version     = "10.0.0"
```
mrueg, line 28: can you rebase and keep the version?

main.go
```diff
@@ -188,6 +188,12 @@ var flags = []cli.Flag{
 		TakesFile: true,
 		EnvVars:   []string{"MARK_INCLUDE_PATH"},
 	}),
+	altsrc.NewStringFlag(&cli.StringFlag{
+		Name:      "update-if-older-than",
+		Value:     "",
+		Usage:     "Update the page only if the last update is older than the specified date.",
```
mrueg, line 194: Can you add the format here in which mark expects the date to be in?

mrueg (commented):
We're currently relicensing to Apache 2.0 only (before it was CommonsClause). Could you mention that you're okay with this being licensed under Apache 2.0?

mrueg:
I think it would also be nice if we could record your snipped `"$(git --no-pager log -1 --pretty="format:%cI" $file)` somewhere in the readme :)

mrueg:
One question/caveat:
If my doc is older and it has a reference to another markdown file which changes the title in a later changeset, there will be an inconsistency I guess.

mrueg:
@danapsimer ping, do you have any interest in continuing on this change?

mrueg:
We use this approach now: https://github.com/kovetskiy/mark/pull/548

PR #497, opened 2024-08-27
Fix error on delete label

### LINKED ISSUE #495: Early exit when deleting Label
## What happened?

When deleting labels, the application ends immediately with a `2024-08-26 16:24:38.278 FATAL EOF` message which interrupts any subsequent operations when globing files.

## What did you expect to happen?

Application can delete labels without errors.

1. Create a page with some labels
2. Remove the label for the mark down 
3. Update the page 
4. Error is generated

* Mark Version (`mark --version`): 10.0.1
* Mark Parameters: n/a
* Confluence Hosting: Cloud
* Confluence Version: Confluence 1000.0.0-cd144b5a79d5
* Environment specific Information: running locally

## Logs or other output

Please provide logs, other kind of output here.

## Additional context

Deleting labels returns a 204 (No-Content) but the underlying library seems to assume that all requests will have a content and tries to parse it. 

The PR to fix the issue in the underlying library https://github.com/kovetskiy/gopencils/pull/3

@kovetskiy

rodcloutier:
Fixes #495 

PR to alllow http status 204 as valid when deleting labels. 

Note: This depends on PR https://github.com/kovetskiy/gopencils/pull/3 and go.mod/go.sum must be updated once merged

### DIFF
--- a/pkg/confluence/api.go
+++ b/pkg/confluence/api.go
@@ -605,7 +605,7 @@ func (api *API) DeletePageLabel(page *PageInfo, label string) (*LabelInfo, error
 		return nil, err
 	}
 
-	if request.Raw.StatusCode != http.StatusOK {
+	if request.Raw.StatusCode != http.StatusOK && request.Raw.StatusCode != http.StatusNoContent {
 		return nil, newErrorStatusNotOK(request)
 	}
 


Dockerfile
```diff
@@ -5,9 +5,12 @@ COPY / .
 RUN make get \
 && make build
 
-FROM chromedp/headless-shell:latest
```
mrueg, line 8: This change seems unrelated, could you please remove it?

Dockerfile
```diff
@@ -5,9 +5,12 @@ COPY / .
...
+FROM chromedp/headless-shell:stable
 RUN apt-get update \
-&& apt-get install --no-install-recommends -qq ca-certificates bash sed git dumb-init \
```
mrueg, line 10: I get the point of pinning it, unfortunately stable and latest are moving targets, so I would like to avoid that it suddenly breaks.

go.mod
```diff
@@ -40,3 +40,5 @@ require (
 	golang.org/x/sys v0.23.0 // indirect
 	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
 )
+
+replace github.com/kovetskiy/gopencils => github.com/petalmd/gopencils v0.0.0-20240827131356-d8904564d2b8
```
mrueg, line 44: Pinged @kovetskiy and it got merged, can you remove the replace? :)

mrueg (commented):
We're currently relicensing to Apache 2.0 only (before it was CommonsClause). Could you mention that you're okay with this being licensed under Apache 2.0?

Dockerfile
rodcloutier: Sorry about this. This was never intended to be pushed in this repo. It is an internal constraint on our side.

Dockerfile
rodcloutier: Sorry about this. This was never intended to be pushed in this repo. It is an internal constraint on our side.

go.mod
rodcloutier: Done

mrueg:
thanks, looks good. Will merge it this Friday after #487 is merged.

Issue #450, opened 2024-06-03
Pages with same title put different parents are conflicting


ChristofferNissen:
## What happened?

Imagine the following pages:

1. Name: "Observability" Parent: "1"
2. Name: "Observability" Parent: "2"

When syncing to Confluence, what happens is that the 2. page does not show up on confluence, and Page 1 is actually Page 2.

## What did you expect to happen?

I expect to be able to have pages with the same Title under different Parents. 

## How can we reproduce the behavior you experienced?

1. Add page with name: "Observability" and parent: "1"
2. Add page with name: "Observability" and parent: "2"
3. Sync to Confluence with Mark
4. Check contents of pages on Confluence

## Further Information (please complete the following information)

* Mark Version (`mark --version`): 9.11.1
* Mark Parameters: `--drop-h1`,  `--title-from-h1`
* Confluence Hosting: Datacenter
* Confluence Version: 8.5.5
* Environment specific Information: Running on Ubuntu latest agents on Azure DevOps

## Logs or other output

I can not provide logs as they might contain sensitive information as it logs the output of our pages. Let me know if interested in anything specific. 

Logs does not show any errors produced, and mark exits with error code 0

mrueg:
I believe pages do need to have a unique name per space and not per parent.

pwlandoll:
That's correct, I've come across this issue before. Since that's a hard limitation from Confluence, there isn't really a way to change that behavior.

A workaround is to make it easier to ensure that each page title is unique to the space. 

Looking at a similar tool, the `confluencebuilder` plugin for Sphinx in Python, there is a [configuration option](https://sphinxcontrib-confluencebuilder.readthedocs.io/en/stable/configuration/#confval-confluence_publish_postfix) that allows you to add a postfix (or prefix) to a page title. Of particular note is the `hash` value that's available, which is a deterministic hash of the space, parent "path", and page title together. That guarantees that even if two pages have the same title before being uploaded, as long as they have different parent "paths" the appended hash value will be different. 

I've opened #509 with a similar implementation that could help avoid this issue.

PR #509, opened 2024-10-01
Add flag to append hash to pages to ensure unique titles

pwlandoll:
This adds a command line flag and environment variable to configure mark to append hash values to page titles. 

This helps solve the issue where multiple pages with the same name, but under different parents, will conflict (#450). 

Another solution could be to allow arbitrary prefix/postfix text, and/or make this computed hash value available as a template variable to be inserted at any point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -788,6 +788,7 @@ GLOBAL OPTIONS:
    --drop-h1, --h1_drop                          don't include the first H1 heading in Confluence output. (default: false) [$MARK_H1_DROP]
    --strip-linebreaks, -L                        remove linebreaks inside of tags, to accomodate non-standard Confluence behavior (default: false) [$MARK_STRIP_LINEBREAK]
    --title-from-h1, --h1_title                   extract page title from a leading H1 heading. If no H1 heading on a page exists, then title must be set in the page metadata. (default: false) [$MARK_H1_TITLE]
+   --title-append-generated-hash                 appends a short hash generated from the path of the page (space, parents, and title) to the title (default: false) [$MARK_TITLE_APPEND_GENERATED_HASH]
    --minor-edit                                  don't send notifications while updating Confluence page. (default: false) [$MARK_MINOR_EDIT]
    --version-message value                       add a message to the page version, to explain the edit (default: "") [$MARK_VERSION_MESSAGE]
    --color value                                 display logs in color. Possible values: auto, never. (default: "auto") [$MARK_COLOR]
--- a/metadata/metadata.go
+++ b/metadata/metadata.go
@@ -3,6 +3,8 @@ package metadata
 import (
 	"bufio"
 	"bytes"
+	"crypto/sha256"
+	"fmt"
 	"regexp"
 	"strings"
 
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -788,6 +788,7 @@ GLOBAL OPTIONS:
    --drop-h1, --h1_drop                          don't include the first H1 heading in Confluence output. (default: false) [$MARK_H1_DROP]
    --strip-linebreaks, -L                        remove linebreaks inside of tags, to accomodate non-standard Confluence behavior (default: false) [$MARK_STRIP_LINEBREAK]
    --title-from-h1, --h1_title                   extract page title from a leading H1 heading. If no H1 heading on a page exists, then title must be set in the page metadata. (default: false) [$MARK_H1_TITLE]
+   --title-append-generated-hash                 append to the title of each page a hash value generated from the page's path (space, parents, and title) (default: false) [$MARK_TITLE_APPEND_GENERATED_HASH]
```
mrueg, line 791:
```suggestion
   --title-append-generated-hash                 appends a short hash generated from the path of the page (space, parents, and title) to the title (default: false) [$MARK_TITLE_APPEND_GENERATED_HASH]
```

metadata/metadata.go
```diff
@@ -164,6 +166,23 @@ func ExtractMeta(data []byte, spaceFromCli string, titleFromH1 bool, parents []s
...
+	// deterministically generate a hash from the page's parents, space, and title
+	if titleAppendGeneratedHash {
+		log.Tracef(
+			nil,
+			"appending hash to page title: %s",
+			meta.Title,
+		)
+		path := strings.Join(append(meta.Parents, meta.Space, meta.Title), "/")
+		pathHash := sha256.Sum256([]byte(path))
+		meta.Title = fmt.Sprintf("%s - %x", meta.Title, pathHash[0:4])
```
mrueg, line 178:
```suggestion
		meta.Title = fmt.Sprintf("%s - %x", meta.Title, pathHash[:7])
```

let's take 7 characters here, just to be safe (in case someone generates a lot of pages)

metadata/metadata.go
```diff
@@ -164,6 +166,23 @@ func ExtractMeta(data []byte, spaceFromCli string, titleFromH1 bool, parents []s
 		meta.Parents = append(parents, meta.Parents...)
 	}
 
+	// deterministically generate a hash from the page's parents, space, and title
+	if titleAppendGeneratedHash {
+		log.Tracef(
```
mrueg, line 171: If you place the log below the hash generation, you can include the hash in the log info, also please use Debugf instead.

metadata/metadata.go
pwlandoll: `pathHash` here is an array of bytes being printed as hexadecimal, so each byte is printed as two characters. The first four bytes and the first eight hexadecimal characters are the same, so I could change it to `[:7]` of the hex string if you think that is more clear. 

Either way, I'll add a comment there since it isn't clear what's happening.

mrueg: Ah I didn't see that it's a byte array and hex representation. This looks good (and a comment is a goo idea). Ignore my suggestion then :)

pwlandoll: I've added a comment for clarity, and included other suggestions.

Thanks for the review!

mrueg:
Thanks for the contribution!

PR #510, opened 2024-10-01
Use query param for label deletion

Smaug123:
(I struggle greatly with Golang, so please assume that if it's possible to do anything wrong, I have done it wrong.)

The Confluence API [is documented](https://developer.atlassian.com/cloud/confluence/rest/v1/api-group-content-labels/#api-wiki-rest-api-content-id-label-label-delete) as preferring this endpoint when the label has slashes in.

I have *not* checked whether it's necessary to URL query-parameter escape this; maybe it is necessary?

(The whitespace change appears to satisfy `go fmt`, but surprised me.)

### DIFF
--- a/confluence/api.go
+++ b/confluence/api.go
@@ -599,8 +599,8 @@ func (api *API) AddPageLabels(page *PageInfo, newLabels []string) (*LabelInfo, e
 func (api *API) DeletePageLabel(page *PageInfo, label string) (*LabelInfo, error) {
 
 	request, err := api.rest.Res(
-		"content/"+page.ID+"/label/"+label, &LabelInfo{},
-	).Delete()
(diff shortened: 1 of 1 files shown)


mrueg:
Hi @Smaug123 the change looks good to me. Have you tested it yourself?

Smaug123:
I only have easy access to a rather ancient Confluence. I'll try and remember to check on Monday what testing I did.

Smaug123:
Yep, just tested this locally against our on-prem Confluence. I used `<!-- Label: origin=foo/bar -->` and deployed, observed that the label was added, then removed that line again, deployed, and observed that the label was removed. As far as my use-case goes, this change is correct.

Issue #316, opened 2023-05-11
Add support for named excerpts


chipbite:
**Is your feature request related to a problem? Please describe.**
Support for named excerpts, which enable a confluence page to define several excerpts exposed for other pages to include.

**Describe the solution you'd like**
The excerpt and excerpt-include macros add name attribute to support named excerpts.

**Describe alternatives you've considered**
Only using one excerpt per page.

Ref: https://github.com/kovetskiy/mark/pull/277

mrueg:
This is fixed in https://github.com/kovetskiy/mark/commit/3d96781f477d02c341d70d9c434b8480d688027d

PR #526, opened 2024-11-05
Feature: Support named excerpts (Cloud only)

YuriyAM:
This pull request adds support for named excerpts (currently available only for Confluence Cloud).

Named excerpts allow users to create multiple excerpts on a single page, each identified by a unique name, making it easier to include specific excerpts from one page to another.

- Resolves feature request #316


mrueg:
Thanks for your contribution!

PR #538, opened 2024-12-12
User friendly blockquote title

### LINKED ISSUE #537: Github alerts MD blockquote conversion to Confluence box generates box title which is not so user friendly
The problem is that currently markdown github alerts style BlockQuote convertion to Confluence is persisting gha alert notation i.e `[!TIP]` as the confluence box title which is not so user friendly

When conversion is done gha alert md annotation is converted in confluence to a human readable title for instance `[!TIP]` will become `Tip`

prokod:
- Instead of the not so nice Github Alerts notations

Closes #537

### DIFF
--- a/renderer/blockquote.go
+++ b/renderer/blockquote.go
@@ -3,6 +3,7 @@ package renderer
 import (
 	"fmt"
 	"regexp"
+	"strings"
 
 	"github.com/yuin/goldmark/ast"
 	"github.com/yuin/goldmark/renderer"
@@ -12,14 +13,16 @@ import (
 
 type ConfluenceBlockQuoteRenderer struct {
 	html.Config
-	LevelMap BlockQuoteLevelMap
+	LevelMap       BlockQuoteLevelMap
+	BlockQuoteNode ast.Node
 }
 
 // NewConfluenceRenderer creates a new instance of the ConfluenceRenderer
 func NewConfluenceBlockQuoteRenderer(opts ...html.Option) renderer.NodeRenderer {
 	return &ConfluenceBlockQuoteRenderer{
-		Config:   html.NewConfig(),
-		LevelMap: nil,
+		Config:         html.NewConfig(),
+		LevelMap:       nil,
+		BlockQuoteNode: nil,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


testdata/quotes.html
```diff
@@ -49,38 +49,38 @@ <h2 id="Simple-Blockquote">Simple Blockquote</h2>
 <h2 id="GH-Alerts-Heading">GH Alerts Heading</h2>
 <h3 id="Note-Type-Alert-Heading">Note Type Alert Heading</h3>
 <ac:structured-macro ac:name="info"><ac:parameter ac:name="icon">true</ac:parameter><ac:rich-text-body>
-<p>[!NOTE]</p>
+<p> Note  </p>
```
mrueg, line 52: Can we remove the leading and trailing whitespaces on this?

renderer/blockquote.go
```diff
@@ -150,7 +165,27 @@ func ParseBlockQuoteType(node ast.Node, source []byte) BlockQuoteType {
...
+}
+
+// Helper function to update the source in place
+func ghaAlertsUpdateSourceInPlace(source []byte, node ast.Node, value string) {
+	leftNode := node.PreviousSibling().(*ast.Text)
+	midNode := node.(*ast.Text)
+	rightNode := node.NextSibling().(*ast.Text)
+
+	title := string(midNode.Value(source))
+	sanitizedTitle := strings.Replace(strings.ToLower(title), "!", "", -1)
```
mrueg, line 178: Instead of doing this on a character base, can we do it on a word base (i.e. look for the existing ghalerts with a [! in front and ] afterwards and then simply replace what you found?

testdata/quotes.html
mrueg: Thinking about it twice, why not remove the whole paragraph? It should show up in Confluence with the proper macro anyway. WDYT @prokod ?

mrueg:
Looking a bit further into it, I wonder if a Transformer might be the best to use: https://codeberg.org/forgejo/forgejo/src/branch/forgejo/modules/markup/markdown/callout/github.go

mwillem97:
When will this feature be released?

PR #548, opened 2025-01-10
add flag for updating on change

### LINKED ISSUE #490: Add option to only update the confluence document if the document is changed.
I have a large number of markdown files that I want to sync with my confluence space, however, I do not want the history to be cluttered with changes every time I execute my deployment script.  

I thought that if I could compare the document generated by `mark` and the current document in confluence.  I have a branch in my fork with some code but it does not work because the HTML that `mark` is sending is not what we get back through using the "content/{id}" endpoint with "expand" including "body,body.storage".   https://github.com/danapsimer/mark/tree/update-if-changed

**Describe alternatives you've considered**
I have another PR that does help this use case using an "upate-if-older-than" logic. See this PR: https://github.com/kovetskiy/mark/pull/489

**Additional context**

kassemsandarusi:
fixes #490

### DIFF
--- a/main.go
+++ b/main.go
@@ -2,9 +2,12 @@ package main
 
 import (
 	"bytes"
+	"crypto/sha1"
+	"encoding/hex"
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"slices"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -190,6 +193,12 @@ var flags = []cli.Flag{
 		TakesFile: true,
 		EnvVars:   []string{"MARK_INCLUDE_PATH"},
 	}),
+	altsrc.NewBoolFlag(&cli.BoolFlag{
+		Name:    "changes-only",
+		Value:   false,
+		Usage:   "Avoids re-uploading pages that haven't changed since the last run.",
+		EnvVars: []string{"CHANGES_ONLY"},
```
mrueg, line 200:
```suggestion
		EnvVars: []string{"MARK_CHANGES_ONLY"},
```

PR #563, opened 2025-02-17
fix: process multiple files without confluence logic

iyilmaz24:
- when the dry-run or compile-only flags are set, allow the processing of multiple files

### DIFF
--- a/main.go
+++ b/main.go
@@ -286,13 +286,15 @@ func RunMark(cCtx *cli.Context) error {
 
 		target := processFile(file, api, cCtx, creds.PageID, creds.Username)
 
-		log.Infof(
-			nil,
-			"page successfully updated: %s",
-			creds.BaseURL+target.Links.Full,
-		)
-
-		fmt.Println(creds.BaseURL + target.Links.Full)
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -286,13 +286,14 @@ func RunMark(cCtx *cli.Context) error {
...
 		target := processFile(file, api, cCtx, creds.PageID, creds.Username)
 
-		log.Infof(
-			nil,
-			"page successfully updated: %s",
-			creds.BaseURL+target.Links.Full,
-		)
-
-		fmt.Println(creds.BaseURL + target.Links.Full)
+		if target != nil { // on dry-run or compile-only, the target is nil
```
mrueg, line 289: can we additionally check for cCtx.Bool("dry-run") and cCtx.Bool("compile-only")?

iyilmaz24: > can we additionally check for cCtx.Bool("dry-run") and cCtx.Bool("compile-only")?

Yup, went ahead and committed this change.

mrueg:
Thanks for the contribution!

Issue #40, opened 2020-11-30
Support for converting Material for MkDocs' Admonitions to Confluence Info Panels


GrantAdkins:
Would be great to allow Material for MkDocs' Admonitions (https://squidfunk.github.io/mkdocs-material/reference/admonitions/) to be converted to Confluence Info Panels (honoring the Info, Note, Success, Warning, Error panel types).

sheldonhull:
I looked into trying to accomplish this with [asciidoc-confluence](https://github.com/asciidoctor/asciidoctor-confluence) in the past when I used it, but it was very complex. From what I can tell in reviewing this, considering I'm more versed with Go, it's setup to make this possible.

Would you consider a contributor guide to adding new macros?

Forgive the length, but since it's on topic of asking for a new feature, I wrote my progress here on what I looked at so it's from a set of fresh eyes. Maybe with a little bit of contributing guidance it would be easy to extend this out.

## The Current Template Structure & Location

1. You are using the storage format, ie the confluence specific xhtml standards documented [here](https://confluence.atlassian.com/doc/confluence-storage-format-790796544.html) [macro-callout](https://confluence.atlassian.com/conf59/info-tip-note-and-warning-macros-792499127.html)
1. The macro logic is contained here: [macro.go](https://github.com/kovetskiy/mark/blob/master/pkg/mark/macro/macro.go)
1. If we contributed a new macro it would be added to the string map here:

https://github.com/kovetskiy/mark/blob/bcf2acb39fd38c99f6c150f71286618f57a9fda5/pkg/mark/stdlib/stdlib.go#L91

with something like:

```go
`ac:callout`: text(
			`<ac:structured-macro ac:name="{{ .Type }}">`,
			`<ac:parameter ac:name="icon">false</ac:parameter>`,
			`<ac:parameter ac:name="title">{{ .CalloutTitle }}</ac:parameter>`,
			`<ac:rich-text-body>`,
		    `<p>`,
		    `{{ .Message | converttohtml }}`, <------- pseduocode from my hugo style experience
		    `</p>`,
		    `</ac:rich-text-body>`,
			`</ac:structured-macro>`,
),
```

1. This would use the text template I provide in my doc like this:

```text
<!-- Macro: info|warning|note etc
     Template: ac:callout
	 CalloutTitle: My Important Callout Title
     Message: Message-->
```

### Quick Side Note On Text Templates

[...]

PR #564, opened 2025-02-18
Support for converting Material for MkDocs' Admonitions to Confluence Info Panels (#40)

PGlesmann:
# Support for Converting Material for MkDocs' Admonitions to Confluence Info Panels (#40)

## Summary  
This PR adds support for converting **Material for MkDocs' Admonitions** into **Confluence Info Panels**. It includes:  
- A **parser** for detecting admonitions in Markdown.  
- A **renderer** for converting them into Confluence's structured macro format.  
- **Unit tests** to ensure correctness.  

## Implementation Details  
- The parser is adapted from [`goldmark-admonitions`](https://github.com/stefanfritsch/goldmark-admonitions).  
- The renderer has been rewritten to generate Confluence-specific output.  

## Testing  
- Added unit tests to verify parsing and rendering behavior.  

## Notes  
- Please review the renderer changes in particular to ensure they align with project standards.  

Its my first contribution to open source ever, so let me know if any improvements are needed 😅

### DIFF
--- a/markdown/markdown.go
+++ b/markdown/markdown.go
@@ -59,7 +59,14 @@ func (c *ConfluenceExtension) Extend(m goldmark.Markdown) {
 		util.Prioritized(crenderer.NewConfluenceImageRenderer(c.Stdlib, c, c.Path), 100),
 		util.Prioritized(crenderer.NewConfluenceParagraphRenderer(), 100),
 		util.Prioritized(crenderer.NewConfluenceLinkRenderer(), 100),
+		util.Prioritized(crenderer.NewConfluenceAdmonitionRenderer(), 100),
 	))
+
+	m.Parser().AddOptions(
+		parser.WithBlockParsers(
+			util.Prioritized(cparser.NewAdmonitionParser(), 100),
+		),
+	)
 
 	m.Parser().AddOptions(parser.WithInlineParsers(
 		// Must be registered with a higher priority than goldmark's linkParser to make sure goldmark doesn't parse
--- a/markdown/markdown_test.go
+++ b/markdown/markdown_test.go
@@ -66,7 +66,7 @@ func TestCompileMarkdownDropH1(t *testing.T) {
 
 	test := assert.New(t)
 
-        testcases, err := filepath.Glob("testdata/*.md")
+	testcases, err := filepath.Glob("testdata/*.md")
 	if err != nil {
 		panic(err)
 	}
@@ -78,7 +78,7 @@ func TestCompileMarkdownDropH1(t *testing.T) {
 		}
 		var variant string
 		switch filename {
-		case "testdata/quotes.md", "testdata/header.md":
+		case "testdata/quotes.md", "testdata/header.md", "testdata/admonitions.md":
 			variant = "-droph1"
 		default:
 			variant = ""
@@ -99,7 +99,7 @@ func TestCompileMarkdownStripNewlines(t *testing.T) {
 
 	test := assert.New(t)
 
-        testcases, err := filepath.Glob("testdata/*.md")
+	testcases, err := filepath.Glob("testdata/*.md")
 	if err != nil {
 		panic(err)
 	}
@@ -110,12 +110,12 @@ func TestCompileMarkdownStripNewlines(t *testing.T) {
 			panic(err)
 		}
 		var variant string
-                switch filename {
-                case "testdata/quotes.md", "testdata/codes.md", "testdata/newlines.md", "testdata/macro-include.md":
-                        variant = "-stripnewlines"
-                default:
-                        variant = ""
-                }
+		switch filename {
+		case "testdata/quotes.md", "testdata/codes.md", "testdata/newlines.md", "testdata/macro-include.md", "testdata/admonitions.md":
+			variant = "-stripnewlines"
+		default:
+			variant = ""
+		}
 
 		markdown, htmlname, html := loadData(t, filename, variant)
 		actual, _ := CompileMarkdown(markdown, lib, filename, "", 1.0, false, true)
--- a/parser/admonition.go
+++ b/parser/admonition.go
@@ -1,0 +1,341 @@
+package parser
+
+import (
+	"bytes"
+	"fmt"
+
+	"math/rand"
+
+	"github.com/yuin/goldmark/ast"
+	"github.com/yuin/goldmark/parser"
+	"github.com/yuin/goldmark/text"
+	"github.com/yuin/goldmark/util"
+)
+
+// A Admonition struct represents a fenced code block of Markdown text.
+type Admonition struct {
+	ast.BaseBlock
+	AdmonitionClass []byte
+	Title           []byte
+}
+
+// Dump implements Node.Dump .
+func (n *Admonition) Dump(source []byte, level int) {
+	ast.DumpHelper(n, source, level, nil, nil)
+}
+
+// KindAdmonition is a NodeKind of the Admonition node.
+var KindAdmonition = ast.NewNodeKind("Admonition")
+
+// Kind implements Node.Kind.
+func (n *Admonition) Kind() ast.NodeKind {
+	return KindAdmonition
+}
+
+// NewAdmonition return a new Admonition node.
+func NewAdmonition() *Admonition {
+	return &Admonition{
+		BaseBlock: ast.BaseBlock{},
+	}
+}
+
+type admonitionParser struct {
+}
+
+var defaultAdmonitionParser = &admonitionParser{}
+
+// NewAdmonitionParser returns a new BlockParser that
+// parses admonition blocks.
+func NewAdmonitionParser() parser.BlockParser {
+	return defaultAdmonitionParser
+}
+
+type admonitionData struct {
+	ID                string   // The ID of the admonition. This enables nested admonitions with indentation
+	char              byte     // Currently, this is always "!"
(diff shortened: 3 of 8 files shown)


mrueg:
Welcome and thanks for your contribution!

When you're saying "Adapted from", what does that mean exactly? 
I'm asking because mark is licensed Apache-2.0, while goldmark-admonitions is MIT licensed. We can use MIT licensed code inside mark, but if you took (a good number is more than 10 lines of code) from the goldmark-admonitions plugin, you should mention this in a comment in the file header, and mention the authors, the source and the license the code is under.

parser/admonition.go
```diff
@@ -0,0 +1,341 @@
...
+
+	"math/rand"
+
+	"github.com/yuin/goldmark/ast"
+	"github.com/yuin/goldmark/parser"
+	"github.com/yuin/goldmark/text"
+	"github.com/yuin/goldmark/util"
+)
+
+// A Admonition struct represents a fenced code block of Markdown text.
```
mrueg, line 15:
```suggestion
// An Admonition struct represents a fenced code block of Markdown text.
```

parser/admonition.go
```diff
@@ -0,0 +1,341 @@
...
+
+func (b *admonitionParser) Open(parent ast.Node, reader text.Reader, pc parser.Context) (ast.Node, parser.State) {
+	line, _ := reader.PeekLine()
+	pos := pc.BlockOffset()
+	if pos < 0 || line[pos] != '!' {
+		return nil, parser.NoChildren
+	}
+	findent := pos
+
+	// currently useless
```
mrueg, line 77: Why is it in here when the comment says it's useless?

parser/admonition.go
```diff
@@ -0,0 +1,341 @@
...
+	admonitionChar := line[pos]
+	i := pos
+	for ; i < len(line) && line[i] == admonitionChar; i++ {
+	}
+	oAdmonitionLength := i - pos
+	if oAdmonitionLength < 3 {
+		return nil, parser.NoChildren
+	}
+
+	// ========================================================================== //
```
mrueg, line 87: Please remove the lines with = they are quite uncommon and distracting.

renderer/admonition.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+		if node.Kind() == ast.KindBlockquote && !entering {
+			admonitionLevel -= 1
+		}
+		return ast.WalkContinue, nil
+	})
+	return AdmonitionLevelMap
+}
+
+// renderBlockQuote will render a BlockQuote
```
mrueg, line 101:
```suggestion
// renderAdmonition will render a Admonition
```

renderer/admonition.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/yuin/goldmark/renderer/html"
+	"github.com/yuin/goldmark/util"
+)
+
+// HeadingAttributeFilter defines attribute names which heading elements can have
+var AdmonitionAttributeFilter = html.GlobalAttributeFilter
+
+// A Renderer struct is an implementation of renderer.NodeRenderer that renders
+// nodes as (X)HTML.
+type ConfluenceAdmonitionRenderer struct {
```
mrueg, line 18: Please replace Admonition with MkDocsAdmonition. There also is https://github.blog/changelog/2023-12-14-new-markdown-extension-alerts-provide-distinctive-styling-for-significant-content/ which have a similar purpose and we likely will have some proper parse for it as some point in time as well.

markdown/markdown.go
```diff
@@ -59,8 +59,15 @@ func (c *ConfluenceExtension) Extend(m goldmark.Markdown) {
 		util.Prioritized(crenderer.NewConfluenceImageRenderer(c.Stdlib, c, c.Path), 100),
 		util.Prioritized(crenderer.NewConfluenceParagraphRenderer(), 100),
 		util.Prioritized(crenderer.NewConfluenceLinkRenderer(), 100),
+		util.Prioritized(crenderer.NewConfluenceAdmonitionRenderer(), 100),
 	))
 
+	m.Parser().AddOptions(
```
mrueg, line 65: We should probably make this an opt-in parser + renderer that can be defined by a command line as it's not standard.
The project mainly tries to support Github Flavored Markdown with a few additional extensions. https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax

markdown/markdown_test.go
```diff
@@ -110,12 +110,12 @@ func TestCompileMarkdownStripNewlines(t *testing.T) {
...
 		}
 		var variant string
-                switch filename {
-                case "testdata/quotes.md", "testdata/codes.md", "testdata/newlines.md", "testdata/macro-include.md":
-                        variant = "-stripnewlines"
-                default:
-                        variant = ""
-                }
+		switch filename {
+		case "testdata/quotes.md", "testdata/codes.md", "testdata/newlines.md", "testdata/macro-include.md", "testdata/admonitions.md":
```
mrueg, line 114: I don't think you need to test extra for stripped lines in this one

markdown/markdown_test.go
```diff
@@ -78,7 +78,7 @@ func TestCompileMarkdownDropH1(t *testing.T) {
 		}
 		var variant string
 		switch filename {
-		case "testdata/quotes.md", "testdata/header.md":
+		case "testdata/quotes.md", "testdata/header.md", "testdata/admonitions.md":
```
mrueg, line 81: I don't think you need to test extra for dropped h1 in this one

mrueg (commented):
I still need to take a closer look at the parser/renderer. This was mainly quick feedback to keep the loop short.

markdown/markdown.go
PGlesmann: whats the best way to do that should I just add a flag and pass the bool into the constructor?

// NewConfluenceRenderer creates a new instance of the ConfluenceRenderer
func NewConfluenceExtension(stdlib *stdlib.Lib, path string, mermaidProvider string, mermaidScale float64, dropFirstH1 bool, stripNewlines bool, mkDocsAdmonitionEnabled bool) *ConfluenceExtension {
	return &ConfluenceExtension{
		// 
               MkDocsAdmontionsEnabled mkDocsAdmonitionEnabled
	}
}

markdown/markdown_test.go
PGlesmann: Hmm, I’m not quite sure I understand this.

For example, when I remove quotes.md and don’t check for stripped lines or the dropped H1 header, the unit test fails.

The same issue occurs with admonitions, the test seems to fail unless I include stripped lines and the dropped H1 header.

I know it’s probably just me misunderstanding how the test cases work, could you help me figure it out?

mrueg: I mean you can drop it from here and remove the -stripped suffixed file. 
The test case is only needed for ones where the stripped lines version and the regular version would differ. We won't need it to test the admonitions I believe. Same applies to the h1 one.

mrueg:
fwiw the failing unit tests are also happening on master. Still need to figure out why it's happening.

markdown/markdown.go
PGlesmann: would be nice if you could come back on this question

PGlesmann:
I rebased it and addressed your points

markdown/markdown_test.go
PGlesmann: The stripped lines version and regular version do differ though. I can change the testcase if you want, but I tried to keep the testcases consistent with those in quotes.md for example. I wanted to maintain uniformity across the testcases

markdown/markdown.go
mrueg: yeah life got a bit busy and this fell through the cracks. I want to add support for https://github.com/terrastruct/d2 diagrams as well and enable them optionally. 

What I would suggest is a:
```
--enabled-extensions
```
CLI flag, which you could use like

```
--enabled-extensions="d2,admonitions,mermaid" 
``` 
and then check the values in this flag and enable the matching parsers/renderers.

markdown/markdown_test.go
mrueg: my bad if there's a difference, then this should be fine. :)

mrueg:
> @mrueg any chance to review this? Then we could move away internally from our fork :-)

I've responded here https://github.com/kovetskiy/mark/pull/564#discussion_r2094616319
I'll need to finish https://github.com/kovetskiy/mark/pull/598 first and then add the optional parser feature.

ChrisRu82:
thx sorry I missed that comment, much appreciated

markdown/markdown.go
mrueg: @PGlesmann I've added a `--features` flag instead, please add this as an "admonitions" feature there.

PGlesmann: thank you, its done

PGlesmann:
I need guidance on the test cases I've added. I included the mkdocsadmonition feature flag in the existing test cases, but I'm wondering if it would make more sense to create a separate test case specifically for mkdocsadmonition functionality.
The concern is that if I create a dedicated test case, I'd need to add logic to ignore/exclude mkdocsadmonition in the other existing tests. Should I keep the current approach where it's integrated into existing tests, or would separating it be better?

PGlesmann:
squashed the commits and rebase onto latest master

mrueg:
Thanks for the contribution!

PR #565, opened 2025-02-19
feat: continue-on-error flag for handling errors while processing multiple files

iyilmaz24:
- Currently: If multiple files are specified, mark starts processing them but exits on the first error - leaving some files untouched
- Proposed: Adding a continue-on-error flag, that does not change above behavior unless set to true, which in that case will process the entire batch of files & log any encountered errors to console.
- Purpose: This should improve user's experience using mark with batch files, as it makes subsequent runs for error'd batches easier

I would like the maintainer's thoughts on the overall idea / PR but also:
- Formerly on some errors, we only logged the error
- Screenshot 2025-02-19 at 3 26 07 PM
- Now in addition, we can also log a helpful string message. I commented out all the changed errors throughout the code and would like to hear thoughts on the new string messages and which ones you dislike or want to replace.

I have also added some tests in a temporary `batch-tests` directory so that the proposed functionality can be tested.
- This command can be used for testing `./mark --compile-only --continue-on-error  -files "**/testdata/batch-tests/*.md"`

Note: there is a pattern where we handle the error with `fatalErrorHandler` and then `return nil` after. I was not able to directly `return fatalErrorHandler(args...)` due to the `processFile` function signature returning `*confluence.PageInfo`. Any sidesteps / alternatives to this are also welcomed (we are able to `return nil` since a pointer can be nil).

Possible further enhancements: Outputting a errors.txt file for users containing the names (and maybe also error messages) of all the error'd files for easy retrial

### DIFF
--- a/error_handler.go
+++ b/error_handler.go
@@ -1,0 +1,35 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/reconquest/pkg/log"
+)
+
+type FatalErrorHandler struct {
+    ContinueOnError bool
+}
+
+func NewErrorHandler(continueOnError bool) *FatalErrorHandler {
+    return &FatalErrorHandler{
+        ContinueOnError: continueOnError,
+    }
+}
+
+func (h *FatalErrorHandler) Handle(err error, format string, args ...interface{}) {
+	errorMesage := fmt.Sprintf(format, args...)
+
+    if err == nil {	
+		if h.ContinueOnError {	
+			log.Error(errorMesage)
+			return
+		}
+		log.Fatal(errorMesage)
+    }
+    
+    if h.ContinueOnError { 
+        log.Errorf(err, errorMesage)
+		return
+    }
+    log.Fatalf(err, errorMesage)
+}
--- a/main.go
+++ b/main.go
@@ -45,6 +45,12 @@ var flags = []cli.Flag{
 		EnvVars:   []string{"MARK_FILES"},
 	}),
 	altsrc.NewBoolFlag(&cli.BoolFlag{
+		Name:    "continue-on-error",
+		Value:   false,
+		Usage:   "dont exit if an error occurs while processing a file, continue processing remaining files.",
+		EnvVars: []string{"MARK_CONTINUE_ON_ERROR"},
+	}),
+	altsrc.NewBoolFlag(&cli.BoolFlag{
 		Name:    "compile-only",
 		Value:   false,
 		Usage:   "show resulting HTML and don't update Confluence page content.",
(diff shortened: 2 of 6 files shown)


main.go
```diff
@@ -329,29 +345,37 @@ func processFile(
...
 
 	if pageID == "" && meta == nil {
-		log.Fatal(
-			`specified file doesn't contain metadata ` +
-				`and URL is not specified via command line ` +
-				`or doesn't contain pageId GET-parameter`,
-		)
+		fatalErrorHandler.Handle(nil, "specified file doesn't contain metadata and URL is not specified via command line or doesn't contain pageId GET-parameter")
+		return nil
+		// log.Fatal(
```
richscott, line 350: Probably this commented-out code should just be removed.

main.go
```diff
@@ -276,6 +282,11 @@ func RunMark(cCtx *cli.Context) error {
 		}
 	}
 
+	fatalErrorHandler := NewErrorHandler(cCtx.Bool("continue-on-error"))
+
+	fmt.Printf("Processing %d files\n", len(files))
```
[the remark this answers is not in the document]
iyilmaz24: I left these in so that pulling the changes and testing them could be easier - definitely will go ahead and remove them 👍

error_handler.go
```diff
@@ -0,0 +1,34 @@
...
+			return
+		}
+		log.Fatal(fmt.Sprintf(format, args...))
+    }
+    
+    if h.ContinueOnError {
+        log.Errorf(err, format, args...)
+        return
+    }
+    log.Fatalf(err, format, args...)
```
mrueg, line 33: This will still be triggered if err == nil, is that intended behavior?

iyilmaz24:
Hi, I was wondering if this was ready to merge - I was looking into creating some batch tests for this functionality. Would it be better to add them to this PR or make a new PR?

mrueg:
feel free to include them here, it would be good if they were executable using the regular tests or bats acceptance tests if necessary.

iyilmaz24:
Hi, the batch tests have been added - with tons of help from @richscott (thanks!)

We also made the decision to refactor/relocate some of the code from the main.go file into a util package. It was necessary to import certain functions and structs for the new batch tests which wasn't possible with the previous code organization. We hope this code relocation also helps in the future with extensibility. Excited to hear any thoughts on the changes 🙂

iyilmaz24:
Hello, I was able to rebase the PR onto the master branch and it should be good to go. 🙂

richscott:
I agree completely about the `util/` directory - it's definitely an example of the "utils folder kitchen-junk-drawer" anti-pattern. I thought about different package topologies, but didn't feel confident about any that would be durably correct, so out of expediency I made `util` for now, so Irfan could get the continue-on-error logic into a unit-test.

richscott:
HI, @mrueg - can you (or anyone else with privileges on this repo) restart these CI checks on this PR? They had the error `The job was not started because recent account payments have failed or your spending limit needs to be increased. Please check the 'Billing & plans' section in your settings` - I assume it's a transient error, maybe?  We'd love to see this PR merged - thanks.

mrueg:
> HI, @mrueg - can you (or anyone else with privileges on this repo) restart these CI checks on this PR? They had the error `The job was not started because recent account payments have failed or your spending limit needs to be increased. Please check the 'Billing & plans' section in your settings` - I assume it's a transient error, maybe? We'd love to see this PR merged - thanks.

I've reached out to @kovetskiy a week ago, hope he find some time to respond.

PR #580, opened 2025-04-04
feat: use gopencils retrial option, upgrade version

iyilmaz24:
This pull request upgrades gopencils and enables its retrial option to reduce connection and reliability issues when connecting to Atlassian Confluence endpoints.


mrueg:
Thanks!

PR #604, opened 2025-06-03
fix: config should be _ instead of -

patrick330602:
config demonstrated on README should be _ instead of -


patrick330602:
> these are aliased, so the hyphenated version should work as well (or did you run into issues here?)

no, in the code they are underscore only and it won't read until I switched from - to _

initially I have trouble using base-url in toml since it never sets until I see this: https://github.com/kovetskiy/mark/blob/master/util/flags.go#L127

which is underline only, which all others are in similar situation

Issue #588, opened 2025-04-23
Support rate limiting


blastur:
If the confluence instance is configured with rate limiting of requests, mark will fail to upload. It would be good if 

The rate limit in confluence is configured as requests per unit of time, per user. So it would be good if these settings could also be passed on the cmdline to mark to make it rate limit requests.

mark --ratelimit 5:3 -f file.md would make it request at most 5 times during 3 seconds.

As a bonus it would be good if mark could understand the error code returned by confluence if it rejects a request due to rate limiting. 

https://developer.atlassian.com/cloud/confluence/rate-limiting/

ppd:
I have a branch here with HTTP retry support, which tackles this exact problem: https://github.com/ppd/mark/tree/http-retry-support

I need to find some time, add tests, and push this upstream.

PR #624, opened 2025-07-20
Back-off and retry mechanism

Yatoom:
Confluence is introducing new [rate limits](https://developer.atlassian.com/cloud/confluence/rate-limiting/). This means that a request sometimes responds with a `429 (Too Many Requests)` status code. However, in contrast to other errors, this is a recoverable error. This pull request implements an exponential back-off and retry mechanism to resolve the problem.

I hope this will be useful. Let me know what you think 👍

### DIFF
--- a/confluence/api.go
+++ b/confluence/api.go
@@ -266,6 +266,7 @@ func (api *API) FindPage(
 	if err != nil {
 		return nil, err
 	}
+
 	if resp.StatusCode == http.StatusTooManyRequests {
 		time.Sleep(1 * time.Second)
 		return api.FindPage(space, title, pageType)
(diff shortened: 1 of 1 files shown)


mrueg:
Thanks for the PR! 
See also: https://github.com/kovetskiy/mark/issues/588
There's a comment from @ppd that suggests to use https://github.com/hashicorp/go-retryablehttp 
This could be a good option as well.

Yatoom:
Oh that looks great! I want to give it a try. Thanks!

mrueg:
https://github.com/kovetskiy/gopencils/pull/5/ has added support for an exponential backoff. Feel free to send a PR to retry on http status code 504 as well.

PR #650, opened 2025-08-29
Add support for using the filename as the page title

dverheijden:
This PR adds support for a `--title-from-filename` flag that will use the filename as the Confluence page name if the title is not explicitly set in the metadata.

## Why this feature is useful
Our usecase of this tool is as follows: We have a "Documentation" repository that houses a lot of markdown files. The organization primarily uses Confluence so, as a compromise, we want to one-way sync our GitHub to Confluence, which this tool can do for us. This way, we can keep doing our documenting in Git, keeping our peer-reviewing and workflows as-is, and still be compliant with company standards.

Our markdown files are mainly RFC-style documents where, usually, the first heading is something like `Introduction` or `Problem statement`. So unless we change every file to have a unique and descriptive first-heading, the `--title-from-h1` doesn't work, since it will overwrite files because of naming collisions. Ideally, we don't notice - nor care - about how the tool syncs it to Confluence. Because we already use a file structure for descriptors of the documents, we would like to use this instead. As such, we think this feature would fit very well.

From a quick adoption POV (like we have), I think this would be the quickest way to adopt this tool for multiple people who just have documents that they want to mirror to Confluence.

### DIFF
--- a/metadata/metadata.go
+++ b/metadata/metadata.go
@@ -159,8 +159,7 @@ func ExtractMeta(data []byte, spaceFromCli string, titleFromH1 bool, titleFromFi
 			meta.Title = ExtractDocumentLeadingH1(data)
 		}
 		if titleFromFilename && meta.Title == "" && filename != "" {
-			base := filepath.Base(filename)
-			meta.Title = strings.TrimSuffix(base, filepath.Ext(base))
+			setTitleFromFilename(meta, filename)
 		}
 		if spaceFromCli != "" && meta.Space == "" {
 			meta.Space = spaceFromCli
@@ -190,6 +189,11 @@ func ExtractMeta(data []byte, spaceFromCli string, titleFromH1 bool, titleFromFi
 	}
 
 	return meta, data[offset:], nil
+}
+
+func setTitleFromFilename(meta *Meta, filename string) {
+	base := filepath.Base(filename)
+	meta.Title = strings.TrimSuffix(base, filepath.Ext(base))
 }
 
 // ExtractDocumentLeadingH1 will extract leading H1 heading
--- a/metadata/metadata.go
+++ b/metadata/metadata.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"crypto/sha256"
 	"fmt"
+	"path/filepath"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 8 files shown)


util/flags.go
```diff
@@ -61,6 +61,12 @@ var Flags = []cli.Flag{
 		Usage:   "extract page title from a leading H1 heading. If no H1 heading on a page exists, then title must be set in the page metadata.",
 		Sources: cli.NewValueSourceChain(cli.EnvVar("MARK_TITLE_FROM_H1"), altsrctoml.TOML("title-from-h1", altsrc.NewStringPtrSourcer(&filename))),
 	},
+	&cli.BoolFlag{
+		Name:    "title-from-filename",
```
mrueg, line 68: what happens if you use --title-from-h1 and --title-from-filename at the same time? i assume these flags should be exclusive.

metadata/metadata.go
```diff
@@ -157,6 +158,10 @@ func ExtractMeta(data []byte, spaceFromCli string, titleFromH1 bool, parents []s
 		if titleFromH1 && meta.Title == "" {
 			meta.Title = ExtractDocumentLeadingH1(data)
 		}
+		if titleFromFilename && meta.Title == "" && filename != "" {
+			base := filepath.Base(filename)
+			meta.Title = strings.TrimSuffix(base, filepath.Ext(base))
```
mrueg: Should this use https://pkg.go.dev/golang.org/x/text/cases#Title and do you think it's useful to replace underscores and with spaces? so `my_new_post.md` becomes `My New Post`

metadata/metadata.go
```diff
@@ -141,7 +142,7 @@ func ExtractMeta(data []byte, spaceFromCli string, titleFromH1 bool, parents []s
 		}
 	}
 
-	if titleFromH1 || spaceFromCli != "" {
+	if titleFromH1 || spaceFromCli != "" || titleFromFilename {
```
mrueg:
```suggestion
	if titleFromH1 || titleFromFilename || spaceFromCli != "" {
```

mrueg:
In general I like the idea, can you also add something to the documentation if you built an order of preference where the title gets taken from.

util/flags.go
dverheijden: Good shout, to avoid confusion I've made them mutually exclusive by adding a CLI check and extended the descriptions

metadata/metadata.go
dverheijden: I've used `cases.Title`. I also changed the existing `strings.Title` while I was at it (since that's currently deprecated).

mrueg:
Thanks for your contribution!

PR #655, opened 2025-09-08
feature: add Server param to Jira ticket template

### LINKED ISSUE #654: Jira template: add `Server` param

Currently, the jira:ticket macro does not allow specifying a server property. This can be limiting when working with multiple Jira instances, since there’s no straightforward way to indicate which server should be used — unlike with jira:filter, which already supports this functionality.

I'd like the jira:ticket macro to support a server parameter, with an implementation similar to the existing support in jira:filter. This would make usage consistent across macros.

**Describe alternatives you've considered**

Using an inline template

**Additional context**

I’d be happy to open a PR implementing this feature if maintainers agree it’s a good idea.

MatyiFKBT:
This pull request makes a small update to the template rendering for Jira ticket macros. 

* Added a new parameter to the Jira macro template to include the server name, using "System JIRA" as a default if `.Server` is not set (`stdlib/stdlib.go`).

Closes #654

### DIFF
--- a/stdlib/stdlib.go
+++ b/stdlib/stdlib.go
@@ -150,6 +150,7 @@ func templates(api *confluence.API) (*template.Template, error) {
 		`ac:jira:ticket`: text(
 			`<ac:structured-macro ac:name="jira">`,
 			`<ac:parameter ac:name="key">{{ .Ticket }}</ac:parameter>`,
+			`<ac:parameter ac:name="server">{{ or .Server "System JIRA" }}</ac:parameter>`,
 			`</ac:structured-macro>`,
 		),
 


stdlib/stdlib.go
```diff
@@ -150,6 +150,7 @@ func templates(api *confluence.API) (*template.Template, error) {
 		`ac:jira:ticket`: text(
 			`<ac:structured-macro ac:name="jira">`,
 			`<ac:parameter ac:name="key">{{ .Ticket }}</ac:parameter>`,
+			`<ac:parameter ac:name="server">{{ or .Server "System JIRA" }}</ac:parameter>`,
```
mrueg: What's the default here? Is it always System JIRA?
Do you also need a ServerID field?

MatyiFKBT: Not sure about the default, we have a custom Jira deployment. Maybe I should not set the default but handle it like this?
```
{{ if .Server  }}
`<ac:parameter ac:name="server">{{ .Server }}</ac:parameter>`
{{ end }}
```

Regarding Server ID, we don't need that, I tested and Confluence can figure out the server only by it's name.

mrueg: thanks for checking, yes let's do it that way.

MatyiFKBT: alright, I updated the code

PR #662, opened 2025-09-29
Strip double quotes from admonition titles

maldoinc:
Mkdocs admonition titles require the quotation marks for it to actually show you that title in the rendered output. However, when this parsed title is passed as-is to confluence, it'll show the panel with the quotes which is both less pleasant but also inconsistent rendering between mkdocs and confluence.

This PR aims to fix that by stripping quotation marks when producing the confluence markup.

### DIFF
--- a/renderer/mkDocsAdmonition.go
+++ b/renderer/mkDocsAdmonition.go
@@ -2,7 +2,7 @@ package renderer
 
 import (
 	"fmt"
-	"strings"
+	"strconv"
 
 	parser "github.com/stefanfritsch/goldmark-admonitions"
 	"github.com/yuin/goldmark/ast"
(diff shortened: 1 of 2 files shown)


renderer/mkDocsAdmonition.go
```diff
@@ -114,8 +115,11 @@ func (r *ConfluenceMkDocsAdmonitionRenderer) renderMkDocsAdmonition(writer util.
 		if _, err := writer.Write([]byte(prefix)); err != nil {
 			return ast.WalkStop, err
 		}
-		if string(n.Title) != "" {
-			titleHTML := fmt.Sprintf("<p><strong>%s</strong></p>\n", string(n.Title))
+
+		title := strings.Trim(string(n.Title), "\"")
```
mrueg: This will likely trim too many quotes. We only want to remove the first and last one, I guess?

mrueg:
Thanks for your contribution! I fixed up the PR to use strconv.Unquote instead.

Issue #690, opened 2025-12-01
Tls verification error


dgudim:
## What happened?

I am trying to use this with our self-hosted confluence, but I'm getting 

```
Get "https://<domain>/rest/api/content/?expand=ancestors%2Cversion&spaceKey=<spacekey>&title=Test&type=page": tls: failed to verify certificate: x509: certificate signed by unknown authority
```

The page opens in the browser just fine and I can `wget` it just fine as well. 

## What did you expect to happen?

No issues with the certificate

## Steps to reproduce the behavior:

1. Try to sync any file

## Further Information

* Mark Version: dev@none
* Mark Parameters: none, only 1 file and config path
* Confluence Hosting: Datacenter
* Confluence Version: 7.7.3
* Environment specific Information: running on ArchLinux

## Additional context

I tried adding the certificate to system trust store explicitly, but that didn't change anything

PR #692, opened 2025-12-05
feature: Add --insecure flag for ignoring tls errors

dgudim:
Closes https://github.com/kovetskiy/mark/issues/690
I am no sure if this is the right way to do this, but it works :)


util/flags.go
```diff
@@ -196,6 +196,12 @@ var Flags = []cli.Flag{
 		Usage:   "Enables optional features. Current features: d2, mermaid, mkdocsadmonitions",
 		Sources: cli.NewValueSourceChain(cli.EnvVar("MARK_FEATURES"), altsrctoml.TOML("features", altsrc.NewStringPtrSourcer(&filename))),
 	},
+	&cli.BoolFlag{
+		Name:    "insecure",
```
mrueg: Can we be explicit and call this argument `insecure-skip-tls-verify`?

dgudim: Yeah, should I change the variable name in `NewAPI` and env var as well?

mrueg: Yes please

dgudim: Here, done. Forgot to update readme, done as well

PR #699, opened 2025-12-17
 Make more genric

fasibio:
Cool work! 
I had the scenario thats an proxy is beween confluence server. 
So therefor some changes happen. 

I did not change logic in general. (expect  **two** places ==> will write comments there) 
Just make it more generic. 

Changes can be collected: 

refactor: 
  - make api rest suffix as flag. 
  - getConfluenceLink make links relativ instand of absolute

feat: 
  - adding flag to extends extra headers

### DIFF
--- a/confluence/api.go
+++ b/confluence/api.go
@@ -98,7 +98,22 @@ func (tracer *tracer) Printf(format string, args ...interface{}) {
 	log.Tracef(nil, tracer.prefix+" "+format, args...)
 }
 
-func NewAPI(baseURL string, username string, password string, insecureSkipVerify bool) *API {
+var _ http.RoundTripper = (*AddingHeaderTransport)(nil)
+
+type AddingHeaderTransport struct {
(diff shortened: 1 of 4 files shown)


page/link.go
```diff
@@ -200,8 +200,7 @@ func getConfluenceLink(
 	space, title string,
 ) (string, error) {
 	link := fmt.Sprintf(
-		"%s/display/%s/%s",
-		api.BaseURL,
+		"/display/%s/%s",
```
fasibio: i think this will be not breaking allways use relativ path

confluence/api.go
```diff
@@ -511,7 +524,7 @@ func (api *API) CreatePage(
 	}
 
 	request, err := api.rest.Res(
-		"content/", &PageInfo{},
+		"content", &PageInfo{},
```
fasibio: i think this will be not breaking but here i am not sure. at my proxy scenario it was necessary.

PR #700, opened 2025-12-18
feat: generate Confluence tiny links for relative markdown links

devantler:
This PR fixes the issue where relative Markdown links like `[DX](./dx/README.md)` were being converted to incorrect Confluence URLs with the `/ex/confluence/<cloudId>/wiki/...` prefix instead of the expected canonical format.

## Solution

Instead of relying on Confluence's `_links.webui` response (which can return the "experience" URL format on Cloud), we now **generate stable Confluence tiny links** directly from the page ID.

**Example:**
- **Before:** `https://example.atlassian.net/ex/confluence/<cloudId>/wiki/spaces/SPACE/pages/12345/PageName`
- **After:** `https://example.atlassian.net/wiki/x/OTAx`

Tiny links are:
- ✅ **Stable** — immune to Cloud-specific URL variations
- ✅ **Universal** — work on both Cloud and Server/DC
- ✅ **Short** — much shorter than full wiki paths
- ✅ **Computed locally** — no additional API calls needed

## Implementation

The tiny link algorithm follows [Atlassian's documented approach](https://support.atlassian.com/confluence/kb/how-to-programmatically-generate-the-tiny-link-of-a-confluence-page/):

1. Convert page ID to little-endian bytes
2. Base64-encode the bytes
3. Apply URL-safe transformations (`/` → `-`, `+` → `_`, strip `=` padding)
4. Format as `{baseURL}/x/{encoded}`

### Changes

**`page/link.go`:**
- `getConfluenceLink()` — now generates tiny links instead of using `_links.webui`
- `GenerateTinyLink(baseURL, pageID)` — exported function for generating tiny links
- `encodeTinyLinkID(id)` — core encoding algorithm

**`page/link_test.go`:**
- `TestEncodeTinyLinkID` — verifies encoding for various page IDs
- `TestGenerateTinyLink` — verifies full URL generation
- `TestEncodeTinyLinkIDMatchesPerl` — validates against Atlassian's documented Perl algorithm
- `TestEncodeTinyLinkIDLargeIDs` — ensures large Cloud page IDs (>32-bit) work correctly

### Additional improvement

The PR also inherits the current document's Confluence space if `--space` is not provided, ensuring relative links are resolved within the correct space.

---

I am unsure if this is the best way to solve [the problem](https://github.com/kovetskiy/mark/issues/701). Please help me understand where to fix my problem if this is not the preferred way to approach it 🙏

Fixes https://github.com/kovetskiy/mark/issues/701

### DIFF
--- a/page/link_test.go
+++ b/page/link_test.go
@@ -74,9 +74,9 @@ func TestEncodeTinyLinkID(t *testing.T) {
 			expected: "EIAB",
 		},
 		{
-			name:     "large page ID from user example",
-			pageID:   5645697027,
-			expected: "A4CCUAE",
+			name:     "large page ID (Confluence Cloud)",
+			pageID:   5000000001,
+			expected: "AfIFKgE",
 		},
 		{
 			name:     "page ID 1",
@@ -114,15 +114,15 @@ func TestGenerateTinyLink(t *testing.T) {
 		{
 			name:     "cloud URL with trailing slash",
 			baseURL:  "https://example.atlassian.net/wiki/",
-			pageID:   "5645697027",
-			expected: "https://example.atlassian.net/wiki/x/A4CCUAE",
+			pageID:   "5000000001",
+			expected: "https://example.atlassian.net/wiki/x/AfIFKgE",
 			wantErr:  false,
 		},
 		{
 			name:     "cloud URL without trailing slash",
 			baseURL:  "https://example.atlassian.net/wiki",
-			pageID:   "5645697027",
-			expected: "https://example.atlassian.net/wiki/x/A4CCUAE",
+			pageID:   "5000000001",
+			expected: "https://example.atlassian.net/wiki/x/AfIFKgE",
 			wantErr:  false,
 		},
 		{
@@ -196,10 +196,10 @@ func TestEncodeTinyLinkIDMatchesPerl(t *testing.T) {
 func TestEncodeTinyLinkIDLargeIDs(t *testing.T) {
 	// Test large page IDs (> 32-bit) which are common in Confluence Cloud
 	// These exceed Perl's pack("L") but our implementation handles them
-	largeID := uint64(5645697027) // User's actual page ID from the issue
+	largeID := uint64(5000000001)
 	result := encodeTinyLinkID(largeID)
 	assert.NotEmpty(t, result)
-	assert.Equal(t, "A4CCUAE", result)
+	assert.Equal(t, "AfIFKgE", result)
 
 	// Verify the result is a valid URL-safe base64-like string
 	assert.Regexp(t, `^[A-Za-z0-9_-]+$`, result)


page/link.go
```diff
@@ -199,28 +206,74 @@ func getConfluenceLink(
...
+	if err != nil {
+		return "", karma.Format(err, "build confluence webui URL")
+	}
+
+	return absolute, nil
+}
 
-	if page != nil {
-		link = api.BaseURL + page.Links.Full
+func makeAbsoluteConfluenceWebUIURL(baseURL string, webui string) (string, error) {
```
mrueg: Should we generate a TinyURL instead? https://support.atlassian.com/confluence/kb/how-to-programmatically-generate-the-tiny-link-of-a-confluence-page/

devantler: Good idea. I gave it a try, and updated the PR to reflect the changes.

devantler: I did use GitHub Copilot with Claude Opus 4.5 for this, as I do not have the necessary context of the code base to do this quick myself. Please be aware of that in the review. 

I did test the changes with Confluence pages on our end, but I am not sure if these changes might introduce issues for other use cases.

```go
cat > /tmp/validate_tiny.go << 'GOEOF'
package main

import (
    "encoding/base64"
    "encoding/binary"
    "fmt"
    "strings"
)

func encodeTinyLinkID(id uint64) string {
    buf := make([]byte, 8)
    binary.LittleEndian.PutUint64(buf, id)
    for len(buf) > 1 && buf[len(buf)-1] == 0 {
        buf = buf[:len(buf)-1]
    }
    encoded := base64.StdEncoding.EncodeToString(buf)
    var result strings.Builder
    for _, c := range encoded {
        switch c {
        case '=':
            continue
        case '/':
            result.WriteByte('-')
        case '+':
            result.WriteByte('_')
        default:
            result.WriteRune(c)
        }
    }
    return result.String()
}

func encodeTinyLinkIDPerl32(id uint32) string {
    buf := make([]byte, 4)
    binary.LittleEndian.PutUint32(buf, id)
    encoded := base64.StdEncoding.EncodeToString(buf)
    var result strings.Builder
    for _, c := range encoded {
        switch c {
        case '=':
            continue
        case '/':
            result.WriteByte('-')
        case '+':
            result.WriteByte('_')
        default:
            result.WriteRune(c)
        }
    }
    s := result.String()
    s = strings.TrimRight(s, "A")
    return s
}

func main() {
    fmt.Println("Comparing Go vs Perl-style (32-bit):")
    testIDs := []uint64{1, 255, 256, 65535, 98319, 98320, 5645697027}
    for _, id := range testIDs {
        goResult := encodeTinyLinkID(id)
        var perlResult string
        if id <= 0xFFFFFFFF {
            perlResult = encodeTinyLinkIDPerl32(uint32(id))
        } else {
            perlResult = "(exceeds 32-bit)"
        }
        match := "OK"
        if goResult != perlResult && perlResult != "(exceeds 32-bit)" {
            match = "MISMATCH"
        }
        fmt.Printf("ID %12d: Go=%-10s Perl32=%-10s %s\n", id, goResult, perlResult, match)
    }
    fmt.Printf("\nUser's page: https://example.atlassian.net/wiki/x/%s\n", encodeTinyLinkID(id))
}
GOEOF
go run /tmp/validate_tiny.go
```

This resulted in a expected URL (Tiny Link format), that I can visit, and which works.

PR #703, opened 2026-01-05
Add math latex support

hypengw:


markdown/markdown.go
```diff
@@ -74,6 +75,7 @@ func (c *ConfluenceExtension) Extend(m goldmark.Markdown) {
 		// Must be registered with a higher priority than goldmark's linkParser to make sure goldmark doesn't parse
 		// the <ac:*/> tags.
 		util.Prioritized(cparser.NewConfluenceTagParser(), 199),
+		util.Prioritized(cparser.NewMathLatexParser(), 200),
```
mrueg, line 78: Can you make this a feature similar to the mkdocsadmonitions above?

renderer/mathlatex.go
```diff
@@ -0,0 +1,58 @@
...
+		w.WriteString("</ac:parameter>")
+		w.WriteString("</ac:structured-macro>")
+	}
+	return ast.WalkContinue, nil
+}
+
+func (r *ConfluenceMathLatexRenderer) renderBlock(w util.BufWriter, source []byte, n ast.Node, entering bool) (ast.WalkStatus, error) {
+	if entering {
+		node := n.(*cparser.MathLatexBlock)
+		quoteType := "ppl mathjax block macro"
```
mrueg, line 48: Is this  a default in confluence or does it need to be installed separately?

hypengw: https://marketplace.atlassian.com/apps/207/latex-and-mathjax?hosting=server&tab=overview

hypengw: Maybe we can make the plugin name a command-line option.
I don’t think generating PNGs for math latex is a good approach.
Another option is to generate HTML with JavaScript resource, but the Confluence environment might be on a private network and unable to access external resources.

mrueg:
This got fixed in https://github.com/kovetskiy/mark/pull/820

Thanks for your contribution though!

PR #709, opened 2026-01-15
add view-file template

hypengw:


hypengw:
https://confluence.atlassian.com/doc/view-file-macro-170494360.html

stdlib/stdlib.go
```diff
@@ -451,6 +451,14 @@ func templates(api *confluence.API) (*template.Template, error) {
 			`<ac:parameter ac:name="autoplay">{{ or .AutoPlay "false"}}</ac:parameter>`,
 			`</ac:structured-macro>`,
 		),
+		`ac:view-file`: text(
```
mrueg, line 455: can you add a link to the template doc as a comment above (see previous templates)?

Issue #717, opened 2026-01-26
Include is not escaped by a code block


ejiektpobehuk:
## What happened?

The include in the following example gets processed: 
````
```
<!-- Include: example.md -->
```
````

So the Confluence page gets a code block with:

````
The contents of the example.md
````

## What did you expect to happen?

The content of a code block is left untouched. So that the Confluence page would show a code block with:

````
<!-- Include: example.md -->
````

## How can we reproduce the behavior you experienced?

1. Create a file to include
2. Create a page that includes the first file in a code block
3. Run mark
4. Check the Confluence page with the include

## Further Information (please complete the following information)

* Mark Version: `mark version 15.3.0@none` from nixpkgs
* Mark Parameters: `--files`
* Confluence Hosting: Server
* Confluence Version: v9.2.13
* Environment specific Information: running in a GitLab CI/CD pipeline

## Logs or other output

Nothing abnormal in the logs

## Additional context

Here is my usecase:

I have a Confluence page dedicate to mark.
Somewhere in the page there is an example of mark heading. In the source code in looks like this:

````
```
<!-- Space: EXMPLSPC -->
<!-- Title: Documentation as code -->
<!-- Include: shared/disclaimer.md -->
```
````

I want my users to copy this part as is.

Unfortunately, only the first two line stay the same while the last one changes to the contents of the `shared/disclaimer.md` file:

```
<!-- Space: EXMPLSPC -->
<!-- Title: Documentation as code -->
!!! warning Disclaimer
  This page is generated from a Markdown file in [${CI_PROJECT_NAME}][docs-repo].
  …

[docs-repo]: ${CI_PROJECT_URL}
```

ejiektpobehuk:
https://github.com/kovetskiy/mark/pull/816 solves the issue

It seems that `<!-- Include: example.md -->` is still going to be expanded but it's out of the scope of this issue

PR #718, opened 2026-01-26
Ignore `Include:` in a ```code block``` for self documentation

ejiektpobehuk:
Addresses #717

## What

`Include` directives inside fenced code blocks are now left untouched, so a Mark-managed page can document Mark's own directive syntax without the examples being expanded.

## How

- Scan the content for all fenced code block ranges (```` ``` ```` and `~~~`,
  per CommonMark §4.5, up to 3 spaces of indentation).
- Find all `Include` directive positions.
- Expand only the directives that fall outside any fenced block; those inside
  are emitted verbatim.

## Before / After

- **Before:** every `Include` directive was processed, including those shown as examples inside code blocks.
- **After:** directives inside fenced code blocks are preserved as-is.

## Scope / limitations

- **Indented (4-space) code blocks are not detected** — directives inside them are still expanded.
- Fences with info strings (e.g. ```` ```bash ````) are handled correctly.

## Notes

`go test ./...` and `go build ./...` pass 

This PR is a part of an internal mark image at my workplace.
We use it to document the way we use mark.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,11 @@ follows the `Include` and `Delims` tag, if present:
      <yaml-data> -->
 ```
 
+**Note:** `Include` directives inside fenced code blocks (```` ``` ```` or
+`~~~`) are left untouched. This makes it possible to document Mark itself
+on a Mark-managed page without the example directives being expanded.
+Directives inside indented (4-space) code blocks are still expanded.
+
 Mark also supports attachments. The standard way involves declaring an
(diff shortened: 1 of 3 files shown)


mrueg:
we switched to goldmark based parsing in #816 so this should not be needed anymore.

ejiektpobehuk:
Great news!

Thanks for letting me know 🙏🏼 

Closing this PR and going to test goldmark based parsing to check if it is time to close related issue.

PR #726, opened 2026-02-16
Add Parent-ID header support

### LINKED ISSUE #523: FR: user page ID instead of page name for publication
Hola again :) 

Currently it's required to use name of the page as destination for publication. For example, if page name is "Test" then we should pass `--parents "test"`.

The problem is that it's not convenient to specify it like that because you need to know name of the page. The problem gets bigger when pages are named not in English language. And imagine if name is too big.

Is it possible to specify ID of the page like numeric ID I see in the URL of non-English-named page? Like `3249049023` in `https://my.org.com/pages/viewpage.action?pageId=3249049023`.

### LINKED ISSUE #644: Page ID
Hi,

Could we consider supporting Confluence page IDs in the  header, alongside existing support for page titles? Relying solely on titles sometimes creates unnecessary fragility in CI/CD workflows, especially when:

	1.	Titles change, which breaks references and fails documentation builds
	2.	Title conflicts occur across different spaces
	3.	Automated processes require stable, immutable references

Using page IDs would provide more reliable parent page resolution and improve long-term maintainability.

Thanks

nobbs:
This PR adds stable parent resolution by Confluence page ID via a dedicated `Parent-ID` header.

## Details
- add `Parent-ID` header support for stable parent resolution by Confluence page ID
- validate that Parent-ID pages belong to the specified space, and warn when Parent headers are ignored
- document the new header and add metadata parsing tests

## Testing
- `go test ./...` (fails in `d2` and `mermaid` image tests; same failures are present on `master` with differing expected dimensions)

## Notes
- `Parent-ID` is header-only (no CLI flag) and is single-use
- when `Parent-ID` is set, `Parent` headers are ignored and logged as a warning
- fixes: #644 
- fixes: #523

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ File in the extended format should follow the specification:
 <!-- Space: <space key> -->
 <!-- Parent: <parent 1> -->
 <!-- Parent: <parent 2> -->
+<!-- Parent-ID: <parent page id> -->
 <!-- Title: <title> -->
 <!-- Attachment: <local path> -->
 <!-- Label: <label 1> -->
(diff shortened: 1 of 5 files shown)


mrueg:
Thanks for the PR. I decided to solve it differently and store that information now in confluence properties. :)

adongy:
Hello! I was interested in something like this but cannot understand how the mentioned pull request helps in the usecase.

Does that mean we should use --track-pages so that the parent folder is tracked across updates? Or is there something usable in the frontmatter properties ? Some help with the updates features would be helpful, thanks!

PR #773, opened 2026-04-24
feat: add "default" as valid Content-Appearance value

moia-sven-ole:
Problem:
When using `<!-- Content-Appearance: fixed -->` the generated Confluence page rendered its body nicely, but the title and page metadata was misaligned with the content.

The page title was left-aligned while the body content was centered, making them visually misaligned.

 Fix:
  The Confluence API uses "default" as the property value for the narrow/fixed-width page layout (what the UI labels "narrow"). Previously, mark only accepted full-width and fixed — but fixed maps to a different Confluence property value
  and causes the misalignment. "default" is the correct value to match what the Confluence UI sets.

  - Adds `"default"` as a valid value for the `Content-Appearance` header and `--content-appearance` CLI flag
  - Passes the value through as-is to the `content-appearance-published` Confluence API property
  - Updates README and adds tests for both the header and CLI flag code paths

  Extends #258 which originally introduced full-width and fixed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,11 +56,12 @@ Also, optional following headers are supported:
 * blogpost: [Blog post](https://confluence.atlassian.com/doc/blog-posts-834222533.html) in `Space`.  Cannot have `Parent`(s)
 
 ```markdown
-<!-- Content-Appearance: (full-width|fixed) -->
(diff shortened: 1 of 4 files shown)


mrueg:
Thanks!

Issue #325, opened 2023-05-26
Support external images in ac:image macro with width property


miikaeru:
I want to keep my repository size as low as possible and thus not upload images to it. This means I have to use external links for adding images to my confluence pages. In addition, I want to add a width to images.

In order to add a width, I would like to use the macro `ac:image`, but it appears not to support external images.

As a workaround, I am using HTML such as
```html
<img alt="Image description" width="700" src="url" />
```

kovetskiy:
could you please provide a reproducible example?

PR #789, opened 2026-06-12
feat: convert <img> HTML tags to ac:image

### LINKED ISSUE #491: HTML <img> Tags Don't Display Images on Confluence
## What happened?

When including something like this:

```markdown
<img src="test.png" width="700" height="100%" style="object-fit:contain;" />
```

The image is not rendered on Confluence.

However, when including this:

```markdown
![](test.png)
```

The image displays correctly on Confluence.

The HTML  tag should correctly display the image on Confluence.

Try syncing a Markdown file to Confluence with the exact HTML  tag.

* Mark Version (`mark --version`): 10.0.1
* Mark Parameters: [e.g. `--drop-h1 --title-from-h1`]
    * -b '{hostname}'
    * --space '{space}'
    * -p $CONFLUENCE_PAT
    * --title-from-h1
    * --drop-h1
    * -f '/docs/docs/**/*.md'
* Confluence Hosting: Server
* Confluence Version: 7.19.25
* Environment specific Information: Docker container used in a Bash script

## Logs or other output

Dry Run Output with Markdown-specific Image Embedding (Works Correctly):

```markdown
<p><ac:image><ri:url ri:value="test.png"/></ac:image></p>
```

Dry Run Output with HTML  Tag (Doesn't Work):

```markdown
<img src="test.png" width="700" height="100%" style="object-fit:contain;"/>
```

---

Confluence Attachment URL with Markdown-specific Image Embedding (Works Correctly):

[https://{hostname}/download/attachments/{id}/._test.png?version=1&modificationDate=1724192742171&api=v2](#)

Confluence Attachment URL with HTML  Tag (Doesn't Work):

[https:/{hostname}/display/{space}/images/test.png](#)

moia-sven-ole:
Feature: Convert  HTML tags to ac:image in Confluence output

Markdown natively supports HTML  tags, and authors commonly use them to control image size while keeping documents readable in standard Markdown renderers like GitHub.
Mark had no handling for these tags, so they would be stripped or passed through as raw HTML, which Confluence does not render.

This change detects  tags in HTML blocks and converts them to Confluence's ac:image format, consistent with how Markdown-syntax images () are already handled. Local files are uploaded as attachments, external URLs are linked directly. The width, alt, and title attributes are all carried over.

The practical benefit over the existing macro-based approach () is that sizing an image no longer requires two separate steps. A single  tag handles the upload and the size in one pass, using syntax that works in many Markdown renderer without mark-specific extensions.

The PR also adds an integration test covering the three main paths (local attachment with attributes, external URL, URL with special characters) and documents the feature in the README.

In addition I think this also fixes https://github.com/kovetskiy/mark/issues/491 and https://github.com/kovetskiy/mark/issues/325

### DIFF
--- a/README.md
+++ b/README.md
@@ -772,6 +772,18 @@ The width will be the commented html after the image (in this case 300px).
 
 Currently this is not compatible with the automated upload of inline images.
 
+### Use HTML img tags for sizing
+
+You can also use standard HTML `<img>` tags on a single line. This allows you to specify sizing while keeping the document readable in standard Markdown renderers like GitHub:
+
+```markdown
+<img src="../images/example.png" width="300" alt="Example" title="An Example" />
+```
+
+Standard attributes (`width`, `alt`, and `title`) are supported and carried over directly. Local image files are uploaded as attachments, and remote image URLs are linked directly.
+
+**Note:** The `<img>` tag must be written on a single line.
+
 ### Render Mermaid Diagram
 
 Confluence doesn't provide [mermaid.js](https://github.com/mermaid-js/mermaid) support natively. Mark provides a convenient way to enable the feature like [GitHub does](https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/).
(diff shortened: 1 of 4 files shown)


raproth:
nice, never has success with marks image width makro. I tried this feature branch, it worked perfectly!

raproth:
what about this PR? I would be very interested to have this merged as I need this feature

mrueg:
Fixed in #826, lives behind a feature gate for now. thanks for your contribution! @moia-sven-ole

PR #798, opened 2026-07-05
fix(confluence): support scoped API tokens via the api.atlassian.com gateway

### LINKED ISSUE #341: mark cli command showing `can't obtain home page from space TEST1`
I am trying to use cli to publish markdown changes to confluence. 
Its giving error 
unable to resolve page
                              └─ can't obtain home page from space "TEST1"
                                 ├─ Confluence API returned unexpected status: 404 (Not Found)
                                 
 ```
<!--Space: TEST1 -->
<!--Parent: On-boarding / Off-boarding -->
<!--Title: GitHub Practice -->
```

The page is already residing in the JIRA, as well as I am using username and password to authenticate, also my base_url is "https://{organisation-name}.atlassian.net/". What am I missing in the above code?

hrabal-myo:
## Problem

Atlassian's **scoped API tokens** only authenticate against the gateway URL `https://api.atlassian.com/ex/confluence/<cloudId>/wiki` — used directly against `<site>.atlassian.net` they are treated as anonymous. Running mark in that setup currently fails in two places:

1. **`FindHomePage` fails with 404** — the v1 endpoint `GET /rest/api/space/<key>` answers 404 for scoped tokens (they lack the classic scopes), so every run aborts with `can't obtain home page from space` (#341 reports this symptom).
2. **Folder support is rejected** — `IsCloud()` only recognizes `*.jira.com` / `*.atlassian.net`, so with the gateway base URL mark refuses `<!-- Folder: -->` even though the gateway always fronts a Cloud instance.

## Fix

- `FindHomePage`: keep the v1 call, but on a non-OK response fall back to the v2 spaces API (`GET /api/v2/spaces?keys=<key>`) and resolve the homepage via `GetPageByID`. Behaviour for classic tokens / Server is unchanged (v1 still wins when it works).
- `IsCloud()`: additionally recognize the host `api.atlassian.com`.

## Testing

- `go build ./...`, `go vet`, `go test ./...` all pass.
- Running in production since 2026-07-03 (as a patch on v16.5.0) in a GitHub Actions pipeline publishing a docs repo to Confluence Cloud with a scoped token via the gateway, including folder parents — pages and folders sync correctly.

Fixes #341

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/confluence/api.go
+++ b/confluence/api.go
@@ -237,15 +237,35 @@ func (api *API) FindHomePage(space string) (*PageInfo, error) {
 	request, err := api.rest.Res(
 		"space/"+space, &SpaceInfo{},
 	).Get(payload)
+	if err == nil && request.Raw.StatusCode == http.StatusOK {
+		return &request.Response.(*SpaceInfo).Homepage, nil
+	}
+
+	// Confluence Cloud answers 404 on the v1 space endpoint for tokens
+	// without classic scopes; fall back to the v2 spaces API.
+	v2Result := struct {
+		Results []struct {
+			ID         string `json:"id"`
+			HomepageID string `json:"homepageId"`
+		} `json:"results"`
+	}{}
+
+	request, err = api.restV2.Res(
+		"spaces", &v2Result,
+	).Get(map[string]string{"keys": space})
 	if err != nil {
-		return nil, err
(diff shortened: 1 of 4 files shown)


Jark:
@hrabal-myo this PR seems 3 weeks old with suggestions from GH copilot on it.

Do you mind if I take a stab at fixing it?

PR #880, opened 2026-08-13
feat(confluence): support the Max page width via Content-Appearance: max

brennanneoh:
Confluence Cloud's editor offers three page widths — Narrow, Wide and Max — and `mark` can currently express only two of them.

Reading `content-appearance-published` / `content-appearance-draft` in Confluence Cloud pages shows the three widths are stored in those same property keys as `default`, `full-width` and `max`. So the editor's Wide is the `full-width` mark already writes, Narrow is `default`, and only `max` is a value mark has no way to produce. Pushing a Max page through `mark` today silently narrows it.

**What this changes**

- `max` is accepted in the `Content-Appearance: ` header and in `--content-appearance`.
- The README documents which editor width each value corresponds to.
- `UpdatePage` writes `content-appearance-draft` alongside `content-appearance-published`.

**About the draft property**

The previous code deliberately skipped the draft key, with a comment referencing [CONFCLOUD-65447](https://jira.atlassian.com/browse/CONFCLOUD-65447) and "sporadic published widths". This deserves a second look.

The reason: writing published alone leaves the two keys disagreeing — draft still says whatever the editor last wrote, published says what mark wrote. The editor takes the draft value, so the width reverts as soon as anyone opens the page. The Confluence editor itself writes both keys with the same value, which is what this now matches. If you'd rather keep the old behavior,  I'll rework it.

**Compatibility**

`full-width`, `fixed` and `default` behave exactly as before, including the implicit `full-width` default when neither the header nor the flag is set.

**Testing**

The fake Confluence server discarded the content properties that ride along on a page update; it now records them, so `TestUpdatePageWritesBothAppearanceProperties` can assert both keys. Existing per-value tests in `util/cli_test.go` and `metadata/metadata_test.go` are extended for `max`.

`go test ./...` passes locally except the `d2` and `mermaid` suites, which need Chrome in `$PATH` and fail the same way on master here. `golangci-lint run ./...` and `markdownlint-cli2` are clean.

The value mapping was verified empirically against a live Confluence Cloud site: a page set to Wide reads back `full-width`, Narrow reads `default`, Max reads `max`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,12 +107,16 @@ Also, optional following headers are supported:
 * blogpost: [Blog post](https://confluence.atlassian.com/doc/blog-posts-834222533.html) in `Space`.  Cannot have `Parent`(s)
 
 ```markdown
-<!-- Content-Appearance: (full-width|fixed|default) -->
+<!-- Content-Appearance: (full-width|fixed|default|max) -->
 ```
 
-* (default) full-width: content will fill the full page width
+The Confluence Cloud editor offers three widths — Narrow, Wide and Max — which
+map onto these values as `default`, `full-width` and `max`.
+
+* (default) full-width: content will fill the full page width. This is the Confluence UI's **Wide**.
 * fixed: content will be rendered in a fixed narrow view
-* default: sets the Confluence property value to `"default"`, which is the narrow layout as set by the Confluence UI. Note: `fixed` maps to a different Confluence property value and can cause misaligned page title and body content — use `default` instead for the narrow layout.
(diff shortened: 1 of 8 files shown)


brennanneoh:
Pushed 88641cf, which fixes a bug I introduced in the first commit.

`ProcessFile` only populates `contentAppearance` when the file carries metadata, so editing a page by id with a metadata-less file reaches `UpdatePage` with an empty appearance. That already wrote an empty `content-appearance-published`; adding the draft key doubled it into the editor-side state, resetting the width of a page nobody asked to restyle. Both keys are now skipped when there is nothing to write, covered by `TestUpdatePageWithoutAppearanceLeavesThePropertiesAlone` (verified failing without the guard).

The commit also restores the CONFCLOUD-65447 reference and the reason the draft key used to be skipped, next to the code that now writes it. My first commit deleted that context along with the behavior, this is on me.

PR #927, opened 2026-09-01
fix: the five ways a publish silently produced the wrong page

mrueg:
Five verified bugs from a code audit, each with a test that fails without the fix. The through-line for the first two: **storage format is XML, and a malformed body makes Confluence reject the whole page** — so every one of these reached the author as an opaque `BadRequestException`, or as nothing at all.

## What is fixed

**Every macro parameter is escaped, in one place.** Fourteen templates interpolated user text straight into the storage format, and `ac:recently-updated` emitted an attribute with *no quotes around it at all* (`stdlib/stdlib.go:478`). A page title containing `&`, or a URL with two query parameters, was enough to break the upload. AGENTS.md invariant 2 has always said to escape through the stdlib funcs — the fix had been applied to `ac:code`, `ac:box`, `ac:status` and the footnotes as each of those broke a page, and never propagated. `xmlesc` now takes `any` so a parameter with a non-string default can pipe through it, and the two call sites that were escaping by hand (`renderer/image.go`, `renderer/math.go`) hand over raw values instead of escaping twice.

**A body Confluence would reject is refused locally.** `CheckWellFormed` parses the finished body — after the layout wrap and after inline comments are merged, since that is what actually gets sent — and fails with the line and the text on it. The parse is the one `renderer/helpers_test.go` has been using; the entity table is HTML's, because the templates emit `&#160;` and documents write `&nbsp;`.

**An admonition marker has to open the line.** `> Information about pricing is below.` published as an `info` macro, and `> Please note that the API changed.` as a `note`, because the classifier matched the word anywhere in the first line. "multiple" contains `tip`; "warning signs" contains `warn`. Everything the documented syntax allows still classifies — `Info: text`, `**Note:**`, `NOTES:`, a bare `Warn`, the `<!-- Info -->` comment form in `testdata/quotes.md`.

**A macro directive inside a code block is an example, not an instruction.** `ExtractMacros` hoisted it out of the fence, registered it as a live macro, and left the fence empty — so mark could not publish its own README, which shows five of them. `findIncludeDirective` and `Macro.Apply` both already skip code regions; only extraction did not.

**Three ways a run lost state**, plus a BOM fix:
- The page manifest was discarded whenever a file failed without `--continue-on-error`. The comment above the save says a single bad file must not throw away the mapping for pages that published fine — but only the continue-on-error branch was fixed. Now deferred, because there are five return paths and the next one added would miss it too.
- `--preserve-comments` silently discarded a retitle: the refetch replaced the page object carrying the staged title. It compounded — the manifest then recorded a title the page did not carry, misresolving every parent named by it.
- `CreateFolder` panicked when a remembered parent folder had been deleted; `GetFolderByID` answers `(nil, nil)`, and this was the only one of its four call sites not checking.
- A UTF-8 BOM made every parser see a file with no metadata, failing with "doesn't contain metadata" — pointing away from the headers, which were present.

## Testing

`go test ./...`, `golangci-lint run ./...` clean. Every fix has a test, and each was confirmed to fail with the fix reverted.

Independent of #925 and #926; all three branch from master.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/confluence/api.go
+++ b/confluence/api.go
@@ -1271,7 +1271,12 @@ func (api *API) CreateFolder(spaceID, title string, parentID *string, parentType
 			return nil, fmt.Errorf("failed to get parent folder info for space consistency: %w", err)
 		}
 
-		if parentFolder.SpaceID != "" {
+		// A folder that is not there answers (nil, nil), not an error, and the
+		// id reaching here came from a cache or from the manifest -- so it names
+		// a folder that existed when it was recorded and may not now. Every
+		// other caller of GetFolderByID checks this; only here did a folder
+		// deleted between runs panic instead of falling back to the space.
+		if parentFolder != nil && parentFolder.SpaceID != "" {
 			actualSpaceID = parentFolder.SpaceID
 		}
 	}
--- a/confluence/folder_test.go
+++ b/confluence/folder_test.go
@@ -1,0 +1,35 @@
+package confluence_test
+
+import (
+	"testing"
+
+	"github.com/kovetskiy/mark/v16/confluence"
+	"github.com/kovetskiy/mark/v16/confluence/confluencetest"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestCreateFolderUnderAVanishedParent: GetFolderByID answers (nil, nil) for a
+// folder that is not there, not an error, and CreateFolder checked only the
+// error before reading the parent's space id -- so it panicked rather than
+// failed.
+//
+// The id reaching it comes from the folder cache or from the manifest, meaning
+// it names a folder that existed when it was recorded. A folder deleted in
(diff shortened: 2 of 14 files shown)


mrueg:
Split into one PR per fix, as requested — easier to review, merge and revert independently:

- #928 `fix(stdlib)`: escape every macro parameter, in one place
- #929 `fix(renderer)`: an admonition marker has to open the line
- #930 `fix(macro)`: a directive inside a code block is an example, not an instruction
- #931 `fix(confluence)`: a parent folder that has vanished is not a panic
- #932 `feat`: refuse a body Confluence would reject, before uploading it
- #933 `fix`: a byte-order mark no longer hides a document's metadata
- #934 `fix`: save the page manifest however the run ends
- #935 `fix`: keep a retitle across the --preserve-comments refetch

Each is independent of the others and branches from master.

PR #1021, opened 2026-09-18
confluence: read _links.base from the create response

### LINKED ISSUE #1019: Links to pages created during the same run are built from --base-url, producing dead api.atlassian.com URLs
## What happened?

Publishing a set of cross-linked pages for the first time produces pages whose inbound links point at the API gateway rather than the site:

```
https://api.atlassian.com/ex/confluence/<cloud-id>/wiki/x/<tiny-id>
```

That URL is an API endpoint, not a browsable page, so every such link 404s for a reader.

The tiny-link IDs themselves are correct — they decode to the right page IDs. Only the host is wrong, and it is wrong for exactly the pages that were **created during the same run**. Links into pages that already existed in Confluence are written correctly.

Against a site-host `--base-url` the bug is invisible, because the fallback happens to be the right value. Behind the `api.atlassian.com` gateway — now the only route for a scoped API token — the fallback is a URL that cannot be opened.

## What did you expect to happen?

That a link to a page mark has just created is the same link mark writes on the next run:

```
https://<tenant>.atlassian.net/wiki/x/<tiny-id>
```

1. Configure mark against the gateway with a scoped API token — `base-url = "https://api.atlassian.com/ex/confluence/<cloud-id>/wiki"`, empty username so the token is sent as a Bearer.
2. Make sure neither `Page A` nor `Page B` exists in the target space yet.
3. Publish both files in one run: `mark --files '*.md'`.
4. Open the published `Page A`. Its link to `Page B` carries the `api.atlassian.com` host and 404s in a browser.
5. Run the same command again with no changes. The link is now `https://<tenant>.atlassian.net/wiki/x/...` and works — because `Page B` now exists and is resolved by a different code path.

In case this is related to specific markdown, please provide a minimal markdown example here.

`a.md`:

```markdown
<!-- Space: SP -->
<!-- Title: Page A -->

# Page A

See [Page B](b.md).
```

`b.md`:

```markdown
<!-- Space: SP -->
<!-- Title: Page B -->

# Page B
```

* Mark Version (`mark --version`): v16.17.2
* Mark Parameters: `--features mermaid --features mention --files '**/*.md' --changes-only --minor-edit --version-message "<sha>" --output-format json`, with `base-url` pointing at the `api.atlassian.com` gateway and an empty username
* Confluence Hosting: Cloud
* Confluence Version: Confluence Cloud
* Environment specific Information: reproduced both on macOS (arm64) and in GitLab CI (`debian:stable-slim`, amd64); credential is a scoped Confluence API token sent as a Bearer

## Logs or other output

Stored XHTML of one page after a first-ever publish of 57 pages, with the links split by whether the target page already existed. Same document, same run:

```
# targets created in an EARLIER run — correct
<a href="https://<tenant>.atlassian.net/wiki/x/KQAnfAE">ADR-0002</a>
<a href="https://<tenant>.atlassian.net/wiki/x/0gAqfAE">ADR-0003</a>

# targets created in the SAME run — dead
<a href="https://api.atlassian.com/ex/confluence/<cloud-id>/wiki/x/RYCefAE">ADR-0018</a>
<a href="https://api.atlassian.com/ex/confluence/<cloud-id>/wiki/x/PoCtfAE">ADR-0027</a>
```

Decoding the tiny-link IDs confirms all four address the intended pages, so only the host prefix differs.

## Additional context

The cause looks to be that only one of the two paths that populate `PageInfo.Links.Base` actually reads it from the response.

`page/link.go:424` takes the base URL from `_links.base` and falls back to `api.BaseURL`:

```go
baseURL := page.Links.Base
if baseURL == "" {
    baseURL = api.BaseURL
}
tiny, err := GenerateTinyLink(baseURL, page.ID)
```

* `FindPage` (`confluence/api.go:638`) reads `_links.base` off the search response and gets the canonical site URL. Correct.
* `CreatePage` (`confluence/api.go:1147`) never reads it — `PageInfo.Links.Base` is tagged `json:"-"` (`confluence/api.go:215`), so the create response's top-level `_links.base` is never unmarshalled — and falls straight through to `page.Links.Base = api.BaseURL`. It then caches the page under that value, so every later lookup in the same run inherits the gateway URL.

**Suggested fix:** have `CreatePage` populate `Links.Base` from the create response the way `FindPage` does — either unmarshal the response's top-level `_links.base` into a local struct and assign it, or drop the `json:"-"` tag if the create response carries the field per-object. Keeping `api.BaseURL` as the last-resort fallback seems right; it just should not win when the server has already said what the canonical URL is.

**Workaround:** run mark twice. The second pass resolves every target through `FindPage`, and `--changes-only` keeps it cheap, since only the pages whose link hosts changed have a different content hash.

Related: #917 — a different symptom of the same gateway, and worth reading together.

team-humaki:
CreatePage dropped the create response's `_links.base` because `PageInfo.Links.Base` was tagged `json:"-"`. Same-run tiny links then fell back to `api.BaseURL`. Behind the `api.atlassian.com` gateway that writes dead API URLs for pages created in the same run; `FindPage` already copies collection `_links.base` and is why the second pass looks right.

The field now unmarshals as `base`. CreatePage still falls back to `api.BaseURL` when the response omits it. Tests cover the tenant URL path, the fallback, and the cache populated by CreatePage.

Fixes #1019


mrueg:
Thanks for your contribution!

